PR #3, opened 2017-12-06
Project initialisation

CameronDiver:


lib/build.ts
```diff
@@ -0,0 +1,99 @@
...
+				resolve(image);
+			});
+		},
+		buildFailure: (error: Error, layers: string[]) => {
+			const image = new LocalImage(
+				docker,
+				layers[layers.length  - 1],
+				task.serviceName,
+				false,
+				false,
```
dfunckt: Booleans are slightly better when passed as an object because it makes clear what they're for.

lib/build.ts
```diff
@@ -0,0 +1,99 @@
...
+			image.error = error;
+
+			resolve(image);
+		},
+		buildStream: (stream: Stream.Duplex) => {
+			if (_.isFunction(task.streamHook)) {
+				task.streamHook(stream);
+			}
+
+			task.buildStream!.pipe(stream);
```
dfunckt: TIL -- you can force unwrap an optional in TS!

lib/external.ts
```diff
@@ -0,0 +1,45 @@
...
+import { DockerProgress } from 'docker-progress';
+import * as Dockerode from 'dockerode';
+import * as _ from 'lodash';
+
+import { BuildTask } from './build-task';
+import { BuildProcessError } from './errors';
+import { LocalImage } from './local-image';
+
+const hasImageTag = (name: string): boolean => {
+	const tagRegex = /.+:[^/]+$/;
```
dfunckt: You need to anchor the regex to the start of string like so: `/^.+:[^/]+$/` otherwise it will match anywhere (which given that the input is not arbitrary is slightly redundant but nice to enforce things where we can).

lib/external.ts
```diff
@@ -0,0 +1,45 @@
...
+	return tagRegex.test(name);
```
dfunckt: This will match a digest which is slightly confusing given the function name.

lib/external.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export function pullExternal(task: BuildTask, docker: Dockerode): Promise<LocalImage> {
+	const dockerProgress = new DockerProgress();
+	// FIXME: Take this out with docker-progress v3.0.1
+	(dockerProgress as any).modem = docker.modem;
```
dfunckt: I think you can remove this now?

lib/external.ts
```diff
@@ -0,0 +1,45 @@
...
+	if (task.imageName == null) {
+		throw new BuildProcessError('No image name given for an external image');
+	}
+	let imageName = task.imageName;
+
+	if (!hasImageTag(imageName)) {
+		imageName += ':latest';
+	}
+
+	// FIXME: resolve images without a tag ( needs to go :latest)
```
dfunckt: I think that's fixed (see above)?

package.json
```diff
@@ -0,0 +1,55 @@
...
+    "@types/mocha": "^2.2.44",
+    "chai": "^4.1.2",
+    "chai-as-promised": "^7.1.1",
+    "gulp": "^3.9.1",
+    "gulp-clean": "^0.3.2",
+    "gulp-sourcemaps": "^2.6.1",
+    "gulp-typescript": "^3.2.3",
+    "gulp-util": "^3.0.8",
+    "husky": "^0.14.3",
+		"mocha": "^4.0.1",
```
dfunckt: Stray tab :)

lib/external.ts
CameronDiver: In docker terms, a digest is also an image tag, this is fine IMO - I can change the naming if you like though?

lib/path-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Given two paths, check whether the first contains the second
+ * @param path1 The potentially containing path
+ * @param path2 The potentially contained path
+ * @return A boolean indicating whether `path1` contains `path2`
+ */
+export const contains = (path1: string, path2: string): boolean => {
+
+	// Fist normalise the input, to remove any path weirdness
```
dfunckt: "Fist" -> "First" :P

package.json
```diff
@@ -0,0 +1,57 @@
...
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/resin-io-modules/resin-multibuild/issues"
+  },
+  "homepage": "https://github.com/resin-io-modules/resin-multibuild#readme",
+  "devDependencies": {
+    "@types/mocha": "^2.2.44",
+    "chai": "^4.1.2",
+    "chai-as-promised": "^7.1.1",
+    "gulp": "^3.9.1",
```
dfunckt: It seems you can remove gulp, related deps and gulpfile.

package.json
```diff
@@ -0,0 +1,57 @@
...
+  "name": "resin-multibuild",
+  "version": "0.0.1",
+  "description": "Build a docker composition",
+  "main": "build/index.js",
+  "types": "lib/index.ts",
+  "scripts": {
+    "prepublish": "npm run build && npm run test",
+    "precommit": "npm run lint && npm run test",
+    "lint": "tslint '{src,typings}/**/*.ts'",
+    "build": "tsc",
```
dfunckt: I think you should be invoking tsc with a project like so `tsc --project ./`.

package.json
CameronDiver: Oh yeah, I was that happy to get circle behaving that I forgot about this haha.

PR #9, opened 2017-12-07
Populate the dockerfile and project type fields of local image

CameronDiver:
Change-type: minor


lib/build.ts
```diff
@@ -30,6 +30,8 @@ function taskHooks(
 				image.layers = layers;
 				image.startTime = startTime;
 				image.endTime = Date.now();
+				image.dockerfile = task.dockerfile;
+				image.projectType = task.projectType || 'no project type available';
```
dfunckt: This is allowed to be null and IMO is better left null.

lib/local-image.ts
```diff
@@ -76,6 +76,12 @@ export class LocalImage {
 	 */
 	public endTime?: number;
 
+	/**
+	 * This is the type of project that was resolved by this module, for
+	 * example Dockerfile.template or Dockerfile.arch.
+	 */
+	public projectType: string;
```
dfunckt: Now I see why you've got the above, but the new SBVR we agreed yesterday allows projectType to be null.

CameronDiver: Ah gotcha, yeah I prefer that.

Issue #10, opened 2017-12-13
Build contexts cannot be shared between services


CameronDiver:
Currently the stream splitting functions assumes only a single matching context, when in reality there can be many.

CameronDiver:
Closed by: https://github.com/resin-io-modules/resin-multibuild/pull/11

PR #11, opened 2017-12-13
Allow a build context to be shared between services

CameronDiver:
This commit changes the stream splitting function to buffer the current
file, and then provide it to all services which use the context that it
is part of.

Change-type: minor
Connects-to: #10
---- Autogenerated Waffleboard Connection: Connects to #10


lib/index.ts
```diff
@@ -45,37 +45,39 @@ export function splitBuildStream(
...
-			const matchingTask = _.find(tasks, (task) => {
+			const matchingTasks = _.filter(tasks, (task) => {
 				if (task.external) {
 					return false;
 				}
 				return PathUtils.contains(task.context!, header.name);
 			});
 
-			if (matchingTask != null) {
+			if (matchingTasks.length === 0) {
```
LucianBuzzo: This can be simply `if (!matchingTasks.length) {`

dfunckt: I would check the happy-path first (i.e. `matchingTasks.length > 0`) like it was before to reduce the diff somewhat.

lib/index.ts
```diff
@@ -45,37 +45,39 @@ export function splitBuildStream(
...
+				Utils.drainStream(stream)
+				.then(() => {
+					next();
+					// return null here to keep bluebird happy
+					return null;
+				})
+				.catch((e) => reject(new TarError(e)));
+			} else {
 				const newHeader = header;
```
zvin: Why do we create a new reference to header here?

CameronDiver: Because we change it on [this line](https://github.com/resin-io-modules/resin-multibuild/pull/11/files#diff-13b5b151431c7e7a17f273559ed212d5R67), not strictly necessary but I believe it shows intent a bit better.

lib/index.ts
CameronDiver: Yeah good idea.

lib/index.ts
zvin: I've seen that but it looks like we think we copy it while we are only creating a reference. It was just a nit anyway.

CameronDiver: Ah I see what you're saying, that could be a bit confusing. Ok you've convinced me, will change.

CameronDiver:
@zvin I don't think that `reject` can be called more than once, because the `next` callback will not be called in any codepath that throws. This means that the outer promise will resolve (well, reject) for the caller and the tar stream handlers will just "stop".

dfunckt:
It's fine to resolve/reject multiple times as well as only the first one will have any effect, i.e. invoking callbacks.

zvin:
I still feel like if an error happens, the build stream will remain "piped" to extract. Is this a bad thing?

CameronDiver:
I don't see this being an issue, but I will make an issue to handle to regardless, it does feel more correct to close the pipe.

Issue #14, opened 2017-12-14
Dont throw an error on project resolution error


CameronDiver:

CameronDiver:
Closed by https://github.com/resin-io-modules/resin-multibuild/pull/15

PR #15, opened 2017-12-14
Indicate project resolution failure with a flag

CameronDiver:
Change-type: minor
Connects-to: #14
---- Autogenerated Waffleboard Connection: Connects to #14


lib/build-task.ts
```diff
@@ -83,4 +83,8 @@ export interface BuildTask {
 	 * the composition
 	 */
 	serviceName: string;
+	/**
+	 * Has this task failed to be resolved?
+	 */
+	resolveFailure: boolean;
```
dfunckt: I think `resolveFailure` makes me expect it's of `Error` type, not a boolean -- let's either store the `ProjectResolutionError` here (if you think that contains useful info) or rename the attribute to `resolved`.

CameronDiver: Yeah it's a bit weird with `BuildTask` being an interface, this need to be defined first, but you're right having it intialised with `false` makes more sense and naming it `resolved`.

lib/resolve.ts
```diff
@@ -44,10 +43,12 @@ export function resolveTask(
 	.then((res: Resolve.ResolvedBundle) => {
 		task.projectType = res.projectType;
 		task.buildStream = res.tarStream;
+		task.resolved = true;
 		return task;
 	})
-	.catch((e: Error) => {
-		throw new ProjectResolutionError(e);
+	.catch(() => {
+		task.projectType = 'Could not be detected';
```
dfunckt: Doesn't need `task.resolved = false;` here as well?

CameronDiver: It's not strictly necessary due to the flag having to be defined from the start: https://github.com/resin-io-modules/resin-multibuild/pull/15/files#diff-37a775e32e4a1ab95c3a6ada682775dcR45 

If you think it makes things clearer though I'm happy to add it in?

CameronDiver:
From a caller's perspective, things can continue as normal, the docker daemon will just throw an error saying that it cannot find a Dockerfile. I guess given this, it's not actually a major change...

lib/resolve.ts
dfunckt: Ah, I see, no then that's fine.

dfunckt:
Yeah I think a minor version bump is fine -- I don't use ProjectResolutionError in the CLI and I think we can be cheeky as I doubt there are users of this lib except the CLI and Builder at the moment.

CameronDiver:
I did think that, I just wasn't sure :p ok yeah I would prefer only a minor bump tbh.

Issue #19, opened 2018-03-13
Support for `dockerfile:` in the `build` section of a compose file


zvin:
We need to support the dockerfile: entry in the build section of a compose file (currently it will only look in the root of a context)

Builder issue: https://github.com/resin-io/resin-builder/issues/420

pdcastro:
This feature ended up being implemented under a different issue:
#34 [Support dockerfile and context directives in docker-compose](https://github.com/balena-io-modules/resin-multibuild/issues/34)

PR #20, opened 2018-03-20
support dockerfile in build

zvin:
---- Autogenerated Waffleboard Connection: Connects to #19


lib/index.ts
```diff
@@ -31,64 +32,82 @@ export * from './local-image';
...
+		extract.on('entry', async (header: tar.TarHeader, stream: Stream.Readable, next: () => void): Promise<void> => {
+			const tasksMatchingAlternateDockerfile = alternateDockerfiles[PathUtils.normalize(header.name)] || [];
+
 			// Find the build context that this file should belong to
 			const matchingTasks = _.filter(tasks, (task) => {
-				if (task.external) {
-					return false;
-				}
-				return PathUtils.contains(task.context!, header.name);
+				return (!task.external || PathUtils.contains(task.context!, header.name));
```
dfunckt: You don’t need the outer parentheses

dfunckt: ignore, for some reason I can’t delete the comment

lib/index.ts
```diff
@@ -31,64 +32,82 @@ export * from './local-image';
...
+		const alternateDockerfiles = getAlternateDockerfiles(tasks);
+
 		const extract = tar.extract();
 
-		const entryFn = (
-			header: tar.TarHeader,
-			stream: Stream.Readable,
-			next: () => void,
-		): void => {
+		extract.on('entry', async (header: tar.TarHeader, stream: Stream.Readable, next: () => void): Promise<void> => {
```
dfunckt: Is typescript adding shims for async/await or compiling them verbatim? I mean, async/await is only in fairly recent Node versions right?

lib/index.ts
```diff
@@ -31,64 +32,82 @@ export * from './local-image';
...
+): Bluebird<BuildTask[]> {
 	const images = Compose.parse(composition);
 	return fromImageDescriptors(images, buildStream);
 }
 
-export function fromImageDescriptors(
-	images: Compose.ImageDescriptor[],
-	buildStream: Stream.Readable,
-): Promise<BuildTask[]> {
+function setDefault(obj: { [s: string]: any }, name: string | number, defaultValue: any): any {
```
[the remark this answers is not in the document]
zvin: TS complains that `An index signature parameter type must be 'string' or 'number'` with the generic key type K so I ended up with:
```typescript 
function setDefault<T>(obj: { [ k: string ]: T }, key: string, defaultValue: T): T {
```

lib/build.ts
```diff
@@ -54,10 +54,12 @@ function taskHooks(
 	};
 }
 
-const generateBuildArgs = (task: BuildTask): { buildargs?: Dict<string> } => {
-	return {
+export const generateBuildArgs = (task: BuildTask): { buildargs?: Dict<string> } => {
```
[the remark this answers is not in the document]
zvin: Updated.

lib/index.ts
zvin: Updated

lib/index.ts
zvin: It depends on the target js version, see https://blog.mariusschulz.com/2016/12/09/typescript-2-1-async-await-for-es3-es5 .
In our case (which is es5 https://github.com/resin-io-modules/resin-multibuild/blob/master/tsconfig.json#L4 ), it will get transpiled. Maybe we should update the target js version as this code will only run in recent node versions?

lib/index.ts
```diff
@@ -31,64 +32,82 @@ export * from './local-image';
...
-): Promise<BuildTask[]> {
+function setDefault<T>(obj: { [ k: string ]: T }, key: string, defaultValue: T): T {
+	// Like Python's `dict.setdefault`
+	const value = obj[key];
+	if (_.isUndefined(value)) {
+		obj[key] = defaultValue;
+		return defaultValue;
+	}
+	return value;
+}
```
CameronDiver: You could do

lib/index.ts
Page-: Bear in mind that await/async has pretty terrible performance in nodejs 6 (and early nodejs 8 versions) @zvin because it disables the v8 optimiser, so we should only really be using the native version if we can rely on nodejs 8.3+ or the code has no need to consider performance at all

lib/index.ts
```diff
@@ -31,64 +32,82 @@ export * from './local-image';
...
 	const images = Compose.parse(composition);
 	return fromImageDescriptors(images, buildStream);
 }
 
-export function fromImageDescriptors(
-	images: Compose.ImageDescriptor[],
-	buildStream: Stream.Readable,
-): Promise<BuildTask[]> {
+function setDefault<T>(obj: { [ k: string ]: T }, key: string, defaultValue: T): T {
+	// Like Python's `dict.setdefault`
```
CameronDiver: I'd rather an explanation about what it actually does :P

lib/index.ts
CameronDiver: Yeah @zvin I'd rather we either update the builder to use a later version of node, or we don't use async/await

lib/index.ts
zvin: Sounds fair , updated :)

lib/index.ts
zvin: @CameronDiver `resin-builder` uses `resin-resin-base:3.3.0` so it has node `8.10.0` https://github.com/resin-io/resin-base/blob/master/Dockerfile#L41 .
What do you think of updating the js target of typescript to `es2017` for this project, `resin-builder` and other things `resin-builder` uses?

CameronDiver: I was not aware that had been updated. In principle that's fine with me, I'd like @Page- 's take on this though.

zvin:
Requested here https://forums.resin.io/t/defining-context-in-docker-compose-error-cannot-locate-specified-dockerfile-dockerfile/2894

I need to double check, but I'm not sure this PR will make Dockerfile *templates* work.

shaunmulligan:
@zvin what is left to get this PR merged?

zvin:
@shaunmulligan this is a bit old now and needs to be rebased.
Also this only supports specifying a specific Dockerfile in the compose file, not a Dockerfile.template.

shaunmulligan:
@zvin and @CameronDiver is it much more work to support Dockerfile.templates? also would it be possible to release it in stages, so regular `Dockerfile` initially and then later add the template resolution?

pdcastro:
This feature ended up being implemented under the following alternative pull requests. A customer has pointed to this PR in the [s/sales Flowdock channel](https://www.flowdock.com/app/rulemotion/s-sales/threads/kB81RDkG7zVrEcgXoi3-TySHE8U) as a feature that needs completing; to avoid other customers getting the same impression, I am taking the liberty of closing this PR (which is also a year old).
* https://github.com/balena-io-modules/resin-multibuild/pull/41
* https://github.com/balena-io-modules/resin-bundle-resolve/pull/27
* https://github.com/balena-io/resin-builder/pull/578

PR #25, opened 2018-03-28
Fix types entry in package.json

CameronDiver:
Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.5.3",
   "description": "Build a docker composition",
   "main": "build/index.js",
-  "types": "lib/index.ts",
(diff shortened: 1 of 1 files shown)


dfunckt:
@CameronDiver this change will make developing tedious -- you'll need to have built the project. Why not change it like it's done on resin-release where it places built files into the same folder? https://github.com/resin-io-modules/resin-release/blob/master/package.json

CameronDiver:
That's what was done, until this: https://github.com/resin-io-modules/resin-multibuild/pull/24

TBh I think I will just revert that, as you make a good point, and I see no reason to add more development tedium.

PR #31, opened 2018-04-30
Export PathUtils module to resin-multibuild importers

CameronDiver:
Change-type: minor


lib/index.ts
```diff
@@ -18,6 +18,7 @@ import * as Utils from './utils';
 export * from './build-task';
 export * from './errors';
 export * from './local-image';
+export { PathUtils };
```
dfunckt: How is this made importable? `import { PathUtils } from 'resin-multibuild`?

CameronDiver: Yeah exactly, or 
```
import * as multibuild from 'resin-multibuild';

multibuild.PathUtils.contains(...)
```

PR #33, opened 2018-09-01
Completely refactor repository

CameronDiver:


.gitignore
```diff
@@ -1 +1,3 @@
 node_modules
+build
+docs
```
hedss: NL

test/mocha.opts
```diff
@@ -0,0 +1 @@
+--timeout 20000
```
hedss: NL

package.json
```diff
@@ -3,14 +3,16 @@
   "version": "0.8.0",
   "description": "Build a docker composition",
   "main": "build/index.js",
-  "types": "lib/index.ts",
+  "types": "build/index.d.ts",
   "scripts": {
+    "clean": "rm -rf build",
```
dfunckt: We usually use `rimraf` which is cross-platform.

tsconfig.publish.json
```diff
@@ -0,0 +1,12 @@
+{
+	"extends": "./tsconfig.json",
```
dfunckt: TIL.

package.json
CameronDiver: Ah, that's nice. I'll add that in in a upcoming PR - thanks!

tsconfig.publish.json
CameronDiver: Yeah it means my testing builds don't get plagued with things like unused locals etc when I'm just trying to develop. It's pretty nice :)

PR #38, opened 2019-01-10
Allow specifying a dockerfile candidate when building

CameronDiver:
It's a major change because now when a file cannot be resolved, it will throw an error (rather than the convoluted setting of a flag).

You can specify a dockerfile, dockerfile.template or any other support dockerfile type in standard docker-compose format.

Change-type: major


lib/resolve.ts
```diff
@@ -38,15 +38,21 @@ export function resolveTask(
...
-		})
-		.catch(() => {
-			task.projectType = 'Could not be detected';
-			return task;
 		});
+		outStream.on('resolved-name', name => {
+			task.dockerfilePath = name;
+		});
+		outStream.on('error', reject);
+		resolve(task);
```
zvin: This may be called before the 'resolver' or 'resolved-name' events. Is this intended?
Also `reject` may be called later.

CameronDiver: The `reject` being called later is definitely an issue actually, but I think that can be handled downstream anyway, the `resolve` should definitely be called first, and these fields are only populated after the build has begun. It's a little convoluted but the only way I could think to get the streaming interface we want.

zvin: It looks like you don't need a Promise at all.
Just add the 'resolver' and 'resolved-name' event listeners.
You can still add task.buildStream.on('error', ...) later.

zvin: or keep the Promise and resolve instead of doing `task.resolved = true`

CameronDiver: Ah you're right, I don't need the promise at all!

PR #40, opened 2019-01-22
Support private registry auth in docker-compose 'image' instruction

pdcastro:
Resolves: #39
Change-type: minor

### DIFF
--- a/lib/external.ts
+++ b/lib/external.ts
@@ -1,4 +1,20 @@
-import * as Promise from 'bluebird';
+/**
+ * @license
+ * Copyright 2019 Balena Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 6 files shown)


lib/registry-secrets.ts
```diff
@@ -89,3 +95,78 @@ export class RegistrySecretValidator {
...
+ *
+ * By example:
+ *   Input: { 'docker.io': { 'username': 'bob',   'password': 'dog' }}
+ *   Output:
+ *   {
+ *     'docker.io': { 'username': 'bob',   'password': 'dog' },
+ *     'https://index.docker.io/v1/': { 'username': 'bob',   'password': 'dog' },
+ *   }
+ */
+export function addCanonicalDockerHubEntry(registryconfig: RegistrySecrets) {
```
pdcastro: This new function will be imported in `resin-builder` in a PR I'm putting together. It will improve the user experience for the Dockerfile `FROM` instruction as well (besides the docker-compose `image:` instruction), in relation to the key (as in key-value pairs) that CLI users are required to use to identify the Docker Hub registry. Currently they are required to use exactly `'https://index.docker.io/v1/'`, but after this change they'll be allowed to use any URL container any of the following 5 domains below, plus the empty string for good measure.

lib/external.ts
```diff
@@ -33,10 +55,19 @@ export function pullExternal(
...
+	if (opts.registryconfig) {
+		authConfigPromise = getAuthConfigObj(imageName, opts.registryconfig);
+	}
 
 	const startTime = Date.now();
-	return dockerProgress
-		.pull(imageName, progressHook, opts)
+	return Bluebird.resolve(authConfigPromise)
+		.then((authConfig: RegistrySecret | {}) => {
+			if (authConfig && Object.keys(authConfig).length > 0) {
```
CameronDiver: We generally use `!_.isEmpty()` in these cases, but it's minor.

PR #42, opened 2019-03-01
Bump version for resin-bundle-resolve

pdcastro:
Change-type: patch


lib/registry-secrets.ts
```diff
@@ -126,6 +126,7 @@ export function addCanonicalDockerHubEntry(registryconfig: RegistrySecrets) {
 		'idx.docker.io',
 		'docker.io',
 		'cloud.docker.com',
+		'hub.docker.com',
```
pdcastro: Also add `'hub.docker.com'` in the list of valid domains for the Docker Hub

PR #48, opened 2019-03-26
Bump docker-progress major version (4.0.0)

pdcastro:
This change allows docker progress stream error events such as `'unauthorized: authentication required'` to be detected.

Change-type: patch

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@types/tar-stream": "1.6.0",
     "ajv": "^6.6.1",
     "bluebird": "^3.5.3",
-    "docker-progress": "^3.0.4",
+    "docker-progress": "^4.0.0",
     "docker-toolbelt": "^3.3.7",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "@types/tar-stream": "1.6.0",
     "ajv": "^6.6.1",
     "bluebird": "^3.5.3",
-    "docker-progress": "^3.0.4",
+    "docker-progress": "^4.0.0",
```
pdcastro: I am reasoning here that although the `docker-progress` change type was 'major' in relation to the "interface" of `docker-progress`, the change is type is 'patch' in relation to the "interface" of `resin-multibuild`. The consequence of this change to users of the `resin-multibuild` package is that some promises will reject in circumstances where they previously fulfilled, but this could be described as a "patch bug fix".

Issue #51, opened 2019-04-16
Fix handling of path separators ('/' vs '\') in tar streams under Windows


pdcastro:
`resin-multibuild` is using the "native" (non POSIX) path handling functions of the Node.js `path` module when parsing tar streams with the [tar-stream](https://www.npmjs.com/package/tar-stream) package. But those streams with that package always use `'/'` as the path separator regardless of the OS/platform, so `resin-multibuild` should use the `path.posix` variant of those functions rather than the native ones.

This is one of the causes of CLI issue: https://github.com/balena-io/balena-cli/issues/1133
There are similar problems with other repos. So far the following have been identified:
* `resin-bundle-resolve`: PR https://github.com/balena-io-modules/resin-bundle-resolve/pull/36
* `node-tar-utils`: PR https://github.com/balena-io-modules/node-tar-utils/pull/8

PR #52, opened 2019-04-17
Fix tar stream path handling under Windows and bump resin-bundle-resolve

pdcastro:
See issue #51 for the main objective of this PR.

User-visible (ish) changes:
* The bump of resin-bundle-resolve provides part of the requirements to allow the use of alternative template or arch-specific Dockerfile names, such as `MyDockerfile.template` or `MyDockerfile.armv7hf`.
* Provides part of the fix for incorrect path handling of tar streams on Windows (forward-slash vs backslash path separators) which causes `balena push ipAddress` and `balena deploy` to fail on Windows.

Resolves: #51
Connects-to: https://github.com/balena-io/balena-cli/issues/1133
Depends-on: https://github.com/balena-io-modules/resin-bundle-resolve/pull/36
Change-type: minor

### DIFF
--- a/lib/path-utils.ts
+++ b/lib/path-utils.ts
@@ -14,24 +14,61 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { normalize, relative } from 'path';
+import * as _ from 'lodash';
+import * as path from 'path';
+import { posix } from 'path';
 
 // Export all of node's path functions, so that users of this
 // module need only import this file
 export * from 'path';
 
+const nativeSepRE = new RegExp(_.escapeRegExp(path.sep), 'g');
+const nativeDotDotRE = new RegExp('^\\.\\.$|\\.\\.' + _.escapeRegExp(path.sep));
+
 /**
- * Given two paths, check whether the first contains the second
- * @param path1 The potentially containing path
- * @param path2 The potentially contained path
+ * Given two native (platform-dependent) paths, check whether the first
+ * contains the second
+ * @param path1 The potentially containing native path
+ * @param path2 The potentially contained native path
  * @return A boolean indicating whether `path1` contains `path2`
  */
 export const contains = (path1: string, path2: string): boolean => {
 	// First normalise the input, to remove any path weirdness
-	path1 = normalize(path1);
-	path2 = normalize(path2);
+	path1 = path.normalize(path1);
+	path2 = path.normalize(path2);
 
 	// Now test if any part of the relative path contains a .. ,
 	// which would tell us that path1 is not part of path2
-	return !/^\.\.$|\.\.\//.test(relative(path1, path2));
+	return !nativeDotDotRE.test(path.relative(path1, path2));
 };
+
+/**
+ * Given two POSIX paths, check whether the first contains the second
+ * @param path1 The potentially containing POSIX path
+ * @param path2 The potentially contained POSIX path
+ * @return A boolean indicating whether `path1` contains `path2`
+ */
+export const posixContains = (path1: string, path2: string): boolean => {
+	// First normalise the input, to remove any path weirdness
+	path1 = posix.normalize(path1);
+	path2 = posix.normalize(path2);
+
+	// Now test if any part of the relative path contains a .. ,
+	// which would tell us that path1 is not part of path2
+	return !/^\.\.$|\.\.\//.test(posix.relative(path1, path2));
+};
+
+/**
+ * Replace "native path separators" like '\' with the POSIX '/' separator.
+ * This function is handy but bear in mind that conversion from Windows paths
+ * to POSIX is not that simple when a drive letter is specified for either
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -48,11 +48,11 @@
     "docker-toolbelt": "^3.3.7",
     "dockerode": "^2.5.8",
     "lodash": "^4.17.4",
-    "resin-bundle-resolve": "^4.0.0",
+    "resin-bundle-resolve": "^4.1.0-35-alternative-template-names-dba0ecea12671ab1793e3d6d3b40ca44418dbb26",
```
pdcastro: Please review the `resin-bundle-resolve` PR so I can update this line... :-)
https://github.com/balena-io-modules/resin-bundle-resolve/pull/36

lib/path-utils.ts
```diff
@@ -14,11 +14,22 @@
...
  * limitations under the License.
  */
-import { normalize, relative } from 'path';
+import { FormatInputPathObject, ParsedPath, posix } from 'path';
 
-// Export all of node's path functions, so that users of this
+// Export all of node's POSIX path functions, so that users of this
 // module need only import this file
-export * from 'path';
+export const basename: (p: string, ext?: string) => string = posix.basename;
```
Page-: These are typed to match the `path` typing right?  In that case it'd be nicer to do
```suggestion
export const basename: typeof path.basename = posix.basename;
```

pdcastro: Actually, in this case there is no point in specifying the type at all, as it is inferred anyway:
```
export const basename = posix.basename;
```
I've done just that.

package.json
```diff
@@ -48,11 +48,11 @@
     "docker-toolbelt": "^3.3.7",
     "dockerode": "^2.5.8",
     "lodash": "^4.17.4",
-    "resin-bundle-resolve": "^4.0.0",
```
pdcastro: Pending review of `resin-bundle-resolve` PR:
https://github.com/balena-io-modules/resin-bundle-resolve/pull/36

lib/path-utils.ts
```diff
@@ -14,11 +14,22 @@
...
+export const dirname = posix.dirname;
+export const extname = posix.extname;
+export const format = posix.format;
+export const join = posix.join;
+export const isAbsolute = posix.isAbsolute;
+export const normalize = posix.normalize;
+export const parse = posix.parse;
+export const relative = posix.relative;
+export const resolve = posix.resolve;
+export const sep = posix.sep;
```
zvin: I think you can write this:
```typescript
export const {
    basename,
    delimiter,
    dirname,
    extname,
    format,
    join,
    isAbsolute,
    normalize,
    parse,
    relative,
    resolve,
    sep,
} = posix;
```

lib/path-utils.ts
```diff
@@ -28,8 +39,8 @@ export * from 'path';
  */
 export const contains = (path1: string, path2: string): boolean => {
 	// First normalise the input, to remove any path weirdness
-	path1 = normalize(path1);
-	path2 = normalize(path2);
+	path1 = posix.normalize(path1);
```
zvin: No need to prefix with `posix.`

lib/path-utils.ts
pdcastro: Yes, it works and it is shorter! Done. 👍

lib/path-utils.ts
pdcastro: Done

lib/path-utils.ts
```diff
@@ -14,24 +14,61 @@
...
 export const contains = (path1: string, path2: string): boolean => {
 	// First normalise the input, to remove any path weirdness
-	path1 = normalize(path1);
-	path2 = normalize(path2);
+	path1 = path.normalize(path1);
+	path2 = path.normalize(path2);
+
+	// Now test if any part of the relative path contains a .. ,
+	// which would tell us that path1 is not part of path2
+	return !nativeDotDotRE.test(path.relative(path1, path2));
```
pdcastro: While this `resin-multibuild` PR is blocked by the [`resin-bundle-resolve`](https://github.com/balena-io-modules/resin-bundle-resolve/pull/36) one... I was working on the CLI and I came across the need for the `path-utils` `contains()` function in native (non-POSIX) flavour. The previous commit in this PR had modified the `contains` function to use POSIX paths (because it is part of the fix for Windows path handling when using the CLI on Windows), but I am now thinking that it would be good to have both native and POSIX versions of `contains` available for use. There are several possibilities on how to refactor `path-utils.ts` and I am not clear that one is much better than the others, and what I've come up with so far was to keep the original `contains` with the same name, and add `posixContains`. Something I did not want to do was to add "if tests" in the body of the `contains` function, because this would likely slow down processing of tar streams.

pdcastro:
Force push: squashed commits and updated `package.json` for version 4.1.0 or `resin-bundle-resolve`.

PR #53, opened 2019-07-04
Support build secrets and variables from a .balena directory

CameronDiver:
Change-type: major


lib/build-metadata.ts
```diff
@@ -0,0 +1,129 @@
...
+				pack.entry(header, buffer);
+			}
+		};
+		return (await TarUtils.cloneTarStream(tarStream, {
+			onEntry,
+		})) as Stream.Readable;
+	}
+
+	public getBalenaYml() {
+		return this.balenaYml;
```
zvin: This allows modifying `this.balenaYml`. Maybe it should return a copy?

lib/build.ts
```diff
@@ -57,13 +58,16 @@ function taskHooks(
...
 }
 
-const generateBuildArgs = (task: BuildTask): { buildargs?: Dict<string> } => {
+const generateBuildArgs = (
+	task: BuildTask,
+	userArgs?: Dictionary<string>,
+): { buildargs?: Dictionary<string> } => {
 	return {
-		buildargs: task.args,
+		buildargs: _.merge(task.args, userArgs),
```
zvin: `_.merge` will mutate `task.args`
```typescript
    buildargs: { ...task.args, ...userArgs}
```

lib/index.ts
```diff
@@ -57,13 +64,19 @@ export function splitBuildStream(
...
 	buildStream: Stream.Readable,
+	metadataDirectory: string = '.balena/',
 ): Promise<BuildTask[]> {
-	return new Promise((resolve, reject) => {
+	// TODO: Get this value from the outside
+	const buildMetadata = new BuildMetadata(metadataDirectory);
+
+	const newStream = await buildMetadata.extractMetadata(buildStream);
+
+	return new Bluebird((resolve, reject) => {
```
zvin: This should probably be a regular `Promise` to match the defined return type.

lib/index.ts
```diff
@@ -119,10 +132,24 @@ export function fromImageDescriptors(
...
+export function buildHasSecrets(tasks: BuildTask[]): boolean {
+	if (tasks.length === 0) {
+		return false;
+	}
+
+	return !_.isEmpty(
+		generateSecretPopulationMap(
+			_.map(tasks, 'serviceName'),
+			tasks[0].buildMetadata,
+			'/tmp',
```
zvin: Maybe `os.tmpdir()` would be better ?

lib/validation-types/varlist.ts
```diff
@@ -0,0 +1,64 @@
...
+};
+
+const convert = (value: unknown): VarList | undefined => {
+	if (!validate(value)) {
+		return;
+	}
+	if (_.isArray(value)) {
+		const varList: VarList = {};
+		_.each(value as string[], str => {
+			const match = str.match(stringRegex);
```
zvin: why not `if (match === null) return;` ?

lib/validation-types/varlist.ts
```diff
@@ -0,0 +1,64 @@
...
+	[key: string]: string;
+}
+
+const stringRegex = /([^\s=]+?)=(\S+)/;
+
+const validate = (value: unknown): value is VarList => {
+	if (_.isArray(value)) {
+		return validateStringArray(value);
+	} else if (_.isObject(value)) {
+		return _.every(value as { [key: string]: unknown }, (v, k) => {
```
zvin: `as Dictionary<unknown>`

lib/build-metadata.ts
CameronDiver: Probably worth doing, as it is though, it only gets used internally.

lib/build.ts
CameronDiver: Ah good catch!

lib/index.ts
CameronDiver: For this it doesn't make a difference, we're just seeing if there would be any secrets to populate. My thoughts for this are that because it won't work on a regular docker daemon, we should surface a way for the caller to detect if build secrets are being used and report that this will only work with balena-engine.

lib/validation-types/varlist.ts
CameronDiver: it's just because due to the validate function above that we know it already exists, but I guess defending against that function changing makes sense, will change.

lib/build-secrets/index.ts
```diff
@@ -0,0 +1,217 @@
...
+				buildSuccess: resolve,
+				buildFailure: reject,
+			});
+		});
+	} catch (e) {
+		throw new SecretPopulationError(e);
+	} finally {
+		// Remove the image
+		try {
+			docker.getImage(imageName).remove({ force: true });
```
zvin: `remove` is async, it should be awaited.

lib/build-secrets/index.ts
```diff
@@ -0,0 +1,217 @@
...
```
zvin: missing await

PR #55, opened 2019-07-12
Fix build secret metadata filtering and filename list

pdcastro:
Change-type: patch

### DIFF
--- a/lib/build-metadata.ts
+++ b/lib/build-metadata.ts
@@ -20,7 +20,7 @@ export class BuildMetadata {
 	private metadataFiles: Dictionary<Buffer> = {};
 	private balenaYml: BalenaYml;
 
-	public constructor(private metadataDirectory: string) {}
+	public constructor(private metadataDirectories: string[]) {}
 
 	public async extractMetadata(
 		tarStream: Stream.Readable,
@@ -59,11 +59,14 @@ export class BuildMetadata {
 	public parseMetadata() {
 		// Yaml takes precedence over json (as our docs are in
 		// yaml), but balena takes precedence over resin
+		// .yml vs .yaml: https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml/
 		const potentials = [
 			{ name: 'balena.yml', type: MetadataFileType.Yaml },
+			{ name: 'balena.yaml', type: MetadataFileType.Yaml },
 			{ name: 'balena.json', type: MetadataFileType.Json },
 			{ name: 'resin.yml', type: MetadataFileType.Yaml },
-			{ name: 'balena.json', type: MetadataFileType.Json },
+			{ name: 'resin.yaml', type: MetadataFileType.Yaml },
+			{ name: 'resin.json', type: MetadataFileType.Json },
 		];
 
 		let bufData: Buffer | undefined;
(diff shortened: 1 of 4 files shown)


lib/build-secrets/index.ts
```diff
@@ -89,7 +89,7 @@ export function generateSecretPopulationMap(
 		secretMap[serviceName] = serviceSecret;
 	}
 
-	return _.omitBy(secretMap, v => _.isEmpty(v));
+	return _.omitBy(secretMap, v => _.isEmpty(v.files));
```
pdcastro: Without this change, services without secret files would fail with the error:
```
...
[Error]   Build failed
SecretPopulationError: The command '/bin/sh -c bash populate.sh' returned a non-zero code: 1
```

lib/build-metadata.ts
```diff
@@ -59,11 +59,14 @@ export class BuildMetadata {
...
 		// yaml), but balena takes precedence over resin
+		// .yml vs .yaml: https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml/
 		const potentials = [
 			{ name: 'balena.yml', type: MetadataFileType.Yaml },
+			{ name: 'balena.yaml', type: MetadataFileType.Yaml },
 			{ name: 'balena.json', type: MetadataFileType.Json },
 			{ name: 'resin.yml', type: MetadataFileType.Yaml },
-			{ name: 'balena.json', type: MetadataFileType.Json },
+			{ name: 'resin.yaml', type: MetadataFileType.Yaml },
+			{ name: 'resin.json', type: MetadataFileType.Json },
```
pdcastro: Note that the changes above are not just adding 'yaml', but also changing `balena.json` to `resin.json`.

lib/index.ts
```diff
@@ -67,9 +67,9 @@ export function splitBuildStream(
 export async function fromImageDescriptors(
 	images: Compose.ImageDescriptor[],
 	buildStream: Stream.Readable,
-	metadataDirectory: string = '.balena/',
+	metadataDirectories = ['.balena/', '.resin/'],
```
pdcastro: `.resin` for compatibility with the older [gist documentation](https://gist.github.com/CameronDiver/6d5964c0618a6154f89d73b883fb09b7) that some customers may still be using in production.

pdcastro:
Force push = adding `export { BalenaYml, ParsedBalenaYml };` to `index.ts`, for use in CLI pull request: https://github.com/balena-io/balena-cli/pull/1344/files

lib/index.ts
```diff
@@ -44,6 +46,7 @@ export * from './build-task';
 export * from './errors';
 export * from './local-image';
 export * from './registry-secrets';
+export { BalenaYml, ParsedBalenaYml };
```
pdcastro: Used in the CLI: https://github.com/balena-io/balena-cli/pull/1344/files

lib/index.ts
CameronDiver: Given that we've never officially supported that, we should try to phase it out

pdcastro: Agreed, it will be easy to change this in a future PR. The existing code before this fix PR supported `resin.yml`, which I thought "goes together" with a metadata directory named `.resin`.

Issue #56, opened 2019-08-19
SecretRemovalError when building on a different arch than the target device


pdcastro:
A recent PR introduced build secret support in `resin-multibuild`. When the docker daemon (that builds the image) runs on a different processor architecture than the target device, `SecretPopulationError` or `SecretRemovalError` are thrown.

Related CLI issues:
* https://github.com/balena-io/balena-cli/issues/1355
* https://github.com/balena-io/balena-cli/issues/1389

PR #57, opened 2019-08-19
Fix SecretRemovalError when building on different arch than the target device

pdcastro:
Resolves: #56 
Change-type: patch

### DIFF
--- a/lib/build-metadata.ts
+++ b/lib/build-metadata.ts
@@ -11,6 +11,8 @@ import { BalenaYml, parsedBalenaYml, ParsedBalenaYml } from './build-secrets';
 import { BalenaYMLValidationError } from './errors';
 import * as PathUtils from './path-utils';
 
+export const QEMU_BIN_NAME = 'qemu-execve';
+
 enum MetadataFileType {
 	Json,
 	Yaml,
@@ -37,10 +39,10 @@ export class BuildMetadata {
 			const buffer = await TarUtils.streamToBuffer(stream);
 			const relative = this.getMetadataRelativePath(header.name);
 
-			if (relative != null) {
-				this.addMetadataFile(relative, buffer);
-			} else {
+			if (relative == null || relative === QEMU_BIN_NAME) {
 				pack.entry(header, buffer);
+			} else {
+				this.addMetadataFile(relative, buffer);
 			}
 		};
 		return (await TarUtils.cloneTarStream(tarStream, {
--- a/lib/build-secrets/Dockerfile
+++ b/lib/build-secrets/Dockerfile
@@ -1,8 +1,8 @@
 #TODO: We'll want this to be packaged into a single image,
 #rather than having to install jq every time for example
-FROM balenalib/%%ARCH%%-alpine:3.9
+FROM %%ARCH%%/alpine:3.10
 
-RUN apk update && apk add jq
+RUN apk update && apk add bash jq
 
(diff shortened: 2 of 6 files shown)


lib/build-secrets/Dockerfile
```diff
@@ -1,8 +1,8 @@
 #TODO: We'll want this to be packaged into a single image,
 #rather than having to install jq every time for example
-FROM balenalib/%%ARCH%%-alpine:3.9
+FROM %%ARCH%%/alpine:3.10
```
pdcastro: During the CLI check-in meeting, Cameron asked me why I changed this from `balenalib` to plain Alpine. The original reason was that I thought that some architectures like `balenalib/amd64-alpine` didn't exist, but I was simply doing something wrong as they do exist. Now, having looked at the it a bit further, a new reason to prefer plain Alpine is the image size: both the base image and final build image (after installing bash and jq) are significantly smaller using plain Alpine:

The base image is less than 6MB for `amd64-alpine:3.10`, and about 35MB for `balenalib/amd64-alpine:3.10`. But the latter already includes `bash` whereas the former doesn't, so I checked the final built size too after installing `bash` and `jq`:
```
$ docker images
REPOSITORY                  TAG         IMAGE ID            CREATED              SIZE
amd64-alpine-bash-310       latest      1fe2fea41ff1        4 seconds ago        12.2MB
blib-amd64-alpine-310       latest      ddc2cb0bb164        About a minute ago   38MB
```
The plain Alpine image at ~12MB is about a third the size of the balenalib alternative at ~38MB.

CameronDiver: This is more than a good enough reason for me :+1: nice work

lib/build-metadata.ts
```diff
@@ -37,10 +39,10 @@ export class BuildMetadata {
 			const buffer = await TarUtils.streamToBuffer(stream);
 			const relative = this.getMetadataRelativePath(header.name);
 
-			if (relative != null) {
-				this.addMetadataFile(relative, buffer);
-			} else {
+			if (relative == null || relative === QEMU_BIN_NAME) {
 				pack.entry(header, buffer);
+			} else {
+				this.addMetadataFile(relative, buffer);
```
pdcastro: I've added a commit to this PR to solve an error I came across while testing the "balena build" command:
```
[Error]   Build failed
Error: COPY failed: stat /var/lib/docker/tmp/docker-builder263845279/.balena/qemu-execve: no such file or directory
```
I reckon that the "metadata extraction" code in resin-multibuild was deleting all contents under the `.balena` metadata folder, including `.balena/qemu-execve` that the CLI adds to images when the `--emulated` flag is given.

@CameronDiver, let me know if you have any concerns about this change, especially the fact that this same code would eventually run in the cloud builders and execute for "balena push" as well.

CameronDiver: @pdcastro this looks fine to me, odd that it didn't trigger every time though :thinking:

Either way, more than happy to get this merged.

pdcastro: I've gathered that this issue was introduced in v11.7.0 - when the metadata code was moved from `resin-builder` to `resin-multibuild`. The error doesn't happen in v11.6.1. I was testing with standalone zip packages for Linux, on an Intel NUC with balenaEngine. Where `SecretRemovalError` applied, however, it would mask the QEMU error.

PR #59, opened 2019-09-03
Add registry secret handling code

CameronDiver:
change-type: minor


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "build/index.js",
   "types": "build/index.d.ts",
   "scripts": {
+    "install": "npm run build",
```
pdcastro: Not this line, but in `package.json`, I think we should bump `resin-bundle-resolve` to 4.1.2 to pickup this PR:
https://github.com/balena-io-modules/resin-bundle-resolve/pull/38
... seeing that `resin-builder`'s own package.json does not mention `resin-bundle-resolve`.

CameronDiver: It would have been picked up regardless but I've made it explicit now

PR #62, opened 2019-09-24
Add support for container contracts

CameronDiver:
We search the root context of a service for a `contract.yml` or `contract.yaml` file. We save these so that the caller can use them however they wish.

We also add some cursory validation, but I plan to add this to the contracts module itself, and integrate that into this module and the supervisor, but didn't want to block on that. In the meantime this will keep things sane.


lib/index.ts
```diff
@@ -101,6 +108,18 @@ export async function fromImageDescriptors(
...
+							// the input stream again to find it
+							if (contracts.isContractFile(relative)) {
+								if (task.contract != null) {
+									throw new MultipleContractsForService(task.serviceName);
+								}
+								task.contract = contracts.processContract(buf);
+							}
+
 							const newHeader = _.cloneDeep(header);
 							newHeader.name = posix.relative(task.context!, header.name);
```
pdcastro: `newHeader.name = relative;`

CameronDiver: Ah yeah, nice catch.

PR #63, opened 2019-09-26
Use pre-built external image for build secret population and removal

pdcastro:
Connects-to: https://github.com/balena-io/balena-cli/issues/1355
Depends-on: https://github.com/balena-io-modules/balena-multibuild-scripts/pull/1
Change-type: patch


lib/build-secrets/index.ts
```diff
@@ -129,14 +126,10 @@ export async function populateSecrets(
...
 	);
-	pack.entry(
-		{ name: 'populate.sh' },
-		await fs.readFile(path.join(__dirname, 'populate.sh')),
-	);
 	const dockerfileContent = dockerfileTemplate.process(
 		await fs.readFile(path.join(__dirname, './Dockerfile'), 'utf8'),
 		{
-			ARCH: getAlpineArch(architecture),
+			ARCH: architecture,
```
pdcastro: The current idea is that the images generated and uploaded to Docker Hub by `balena-io-modules/balena-multibuild-scripts` will be prefixed with the architecture values returned by `docker.version().Arch` (i.e. one of 'arm', 'arm64', '386', 'amd64'), so there's no need for the `getAlpineArch()` logic.

lib/build-secrets/Dockerfile
```diff
@@ -1,8 +1,3 @@
-#TODO: We'll want this to be packaged into a single image,
-#rather than having to install jq every time for example
-FROM %%ARCH%%/alpine:3.10
-
-RUN apk update && apk add bash jq
-
-COPY secrets.json populate.sh ./
-RUN bash populate.sh
+FROM u32915/%%ARCH%%-balena-multibuild-scripts
```
pdcastro: * `u32915` is a personal DockerHub account that I've been using for testing, while the new `balena-io-modules/balena-multibuild-scripts` repo is being setup: https://github.com/balena-io-modules/balena-multibuild-scripts/pull/1
* In the end, depending on how `balena-io-modules/balena-multibuild-scripts` gets setup, the available Docker images may be:
  * `balena/arm-balena-multibuild-scripts`
  * `balena/arm64-balena-multibuild-scripts`
  * `balena/amd64-balena-multibuild-scripts`
  * `balena/386-balena-multibuild-scripts`

lib/build-secrets/populate.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/bash
-set -e
-
-while read -r name content; do
-	mkdir -p $(dirname "$name")
-	echo  "$content" | base64 -d > "$name";
-done <<< $(jq --raw-output '. as $top | keys[] as $k | .[$k].files | keys[] as $f | "\($top[$k].tmpDirectory + "/" + $f) \(.[$f])"' secrets.json)
-
```
pdcastro: `populate.sh` and `remove.sh` were moved to `balena-io-modules/balena-multibuild-scripts`.

lib/index.ts
```diff
@@ -162,20 +162,6 @@ export async function fromImageDescriptors(
 	});
 }
 
-export function buildHasSecrets(tasks: BuildTask[]): boolean {
```
pdcastro: This function was not being used/called anywhere.

test/build-utils.ts
```diff
@@ -0,0 +1,97 @@
+/**
```
pdcastro: This file is used by test code only. The contents were mostly moved here from `build.spec.ts`, so that it could be reused / shared with the new `multibuild.spec.ts`.

test/mocha.opts
```diff
@@ -1,2 +1,2 @@
---timeout 20000
+--timeout 25000
```
pdcastro: When running `multibuild.spec.ts` against a NUC and an RPi3 (to test build secrets and multiple architectures), the RPi3 needed an extra 5 seconds...

test/multibuild.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+class StreamOutputParser {
+	public outputLines: string[] = [];
+
+	public captureOutput(stream: Stream.Readable) {
+		stream.on('data', data => this.outputLines.push(data.toString()));
+	}
+}
+
+describe('performBuilds()', () => {
+	it.skip('correctly builds a task using build secrets', async () => {
```
pdcastro: After writing this test, it's disappointing that I've had to skip it (`it.skip`). I initially forgot that balenaEngine would be required instead of standard Docker. Still, it was useful to run it manually against local balena devices.

lib/index.ts
CameronDiver: It's used in the CLI I think, or at least it should be, so we can avoid pulling the secret population images when they're not needed.

pdcastro: I've restored `buildHasSecrets()` (no problem), but there are no references to it in `balena-cli`, `resin-builder` or `resin-multibuild`, neither before nor after this PR. To avoid pulling the secret population images when they're not needed, a separate test is made in `performBuilds()`, taking advantage of the fact that it already calls `generateSecretPopulationMap()`.

pdcastro:
Waiting for the Docker repos (`balena/*-balena-multibuild-scripts`) to be made public before merging this PR: https://www.flowdock.com/app/rulemotion/r-operations/threads/WWkULxQ9XBtcwJXhrivmgv1kSDb

lib/index.ts
CameronDiver: Hmm I'd say just make a judgement call then @pdcastro, either remove this or change the CLI to only use this function (I think the latter is my preference but I'm easy).

pdcastro:
@balena-ci retest

Issue #64, opened 2019-10-07
"balena build" on Windows: "\\var\\lib\\docker\\tmp\\ed1f9f5a84d092cf7c3e" includes invalid characters for a local volume name


pdcastro:
It appears that the build secrets code moved from `resin-builder` to `resin-multibuild` in the recent past included `path.join`, `path.relative` and similar platform-specific path operations that were actually being applied to the handling of tar streams, which are platform-agnostic (tar streams always use forward slashes as the path separator). As a result, when the code was executed on Windows by `balena-cli`, backslashes were being inserted in the tar stream sent to the Docker daemon in balena devices.
```
C:\balena-cli>node bin\balena build other\simple_container -a test-rpi --nocache --logs --emulated -h 192.168.0.21 -p 2375 -R secrets.yml
[Info]    Creating default composition with source: other\simple_container
[Info]    Building for armv7hf/raspberrypi3
[Build]   Building services...
[Build]   main Preparing...
[Info]    Emulation is enabled
qemuPathInContext is ".balena\qemu-execve"
transposing tar stream
[Build]   main Step 1/7 : FROM eu.gcr.io/buoyant-idea-226013/arm32v7/busybox
[Build]   main  ---> 7ab81f29afdb
[Build]   main Step 2/7 : COPY [".balena/qemu-execve","/tmp/qemu-execve"]
[Build]   main  ---> c6822de8a289
[Build]   Built 1 service in 0:03
[Error]   Build failed
create \var\lib\docker\tmp\ed1f9f5a84d092cf7c3e: "\\var\\lib\\docker\\tmp\\ed1f9f5a84d092cf7c3e" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed. If you intended to pass a host directory, use absolute path
```

pdcastro:
Resolved by #65

PR #65, opened 2019-10-07
Fix path operations and get 'npm test' to pass on Windows

pdcastro:
Addresses issue #64 ["balena build" on Windows: "\\\\var\\\\lib\\\\docker\\\\tmp\\\\ed1f9f5a84d092cf7c3e" includes invalid characters for a local volume name](https://github.com/balena-io-modules/resin-multibuild/issues/64)

Change-type: patch

### DIFF
--- a/lib/build-metadata.ts
+++ b/lib/build-metadata.ts
@@ -81,7 +81,7 @@ export class BuildMetadata {
 	}
 
 	public getSecretFile(source: string): Buffer | undefined {
-		return this.metadataFiles[PathUtils.join('secrets', source)];
+		return this.metadataFiles[PathUtils.posix.join('secrets', source)];
 	}
 
 	public parseMetadata() {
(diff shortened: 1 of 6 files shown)


lib/build-metadata.ts
```diff
@@ -202,9 +202,9 @@ export class BuildMetadata {
 		path: string,
 	): { relativePath: string; metadataDirectory: string } | undefined {
 		for (const metadataDirectory of this.metadataDirectories) {
-			if (PathUtils.contains(metadataDirectory, path)) {
+			if (PathUtils.posixContains(metadataDirectory, path)) {
```
dfunckt: Can you please change `PathUtils` so that `contains` is part of its `posix` namespace? All the other commands are `PathUtils.posix.<command>` but this is on `PathUtils` which is inconsistent. It doesn't have to be on this PR if you don't feel like though.

pdcastro: Whether or not this is a good reason, the reason is that `PathUtils.posix.*` is a plain re-export of the standard Node `path.posix.*` (built-in lib), whereas `PathUtils.contains` and `PathUtils.posixContains` are our own functions. In other words, there is currently consistency in that neither `path.contains` nor `path.posix.contains` exist. :-)
```
# resin-multibuild/lib/path-utils.ts

import { posix } from 'path';
export * from 'path';

export const contains = ... ;
export const posixContains = ... ;
```

pdcastro: So I think I'll leave the `posixContains` → `posix.contains` change to a future PR.

PR #79, opened 2020-06-12
Revert "Propagate target platform to builds"

robertgzr:
This reverts commit 9e75147cfd1a11d0c57d8589a5bc8bd9c5ba43c2.

After looking at the code patch in the builder it turns out that the
dockerOpts we set here get passed back and are modified again which
eventually drops the platform property introduced in the original
commit.

Change-type: minor


robertgzr:
@balena-ci I self-certify!

PR #80, opened 2020-06-30
Resolve docker platform

robertgzr:
Paritally reverts commit b5633196b5ea2d7d9e159024ab4620407ef02d53.

This ensures that the build resolves the correct platform for base
images it fetches from the registry.
Includes unit tests to check the platform translation and the actual
build execution.

This requires running a docker daemon >=18.09.3 and experimental
features enabled, to support this tests will only run if a daemon
matching this requirement is available.

Change-type: minor


robertgzr:
@kaisoz I fixed the issues with my commits. No idea where the other failures come form, any idea?

robertgzr:
@Page- could you take a look here please? I'm getting type errors from some parts of the code I didn't even touch. I suspect something in js-yaml (since the errors happen around parts that work with return values from `jsYaml.safeLoad`)

Page-:
Try https://github.com/balena-io-modules/resin-multibuild/pull/81 @robertgzr

kaisoz:
@balena-ci retest

lib/registry-secrets.ts
```diff
@@ -74,14 +74,15 @@ export class RegistrySecretValidator {
...
+	public validateRegistrySecrets(
+		parsedJson: unknown,
+	): asserts parsedJson is RegistrySecrets {
 		const valid = this.validateFunction(parsedJson);
 		if (!valid) {
 			throw new RegistrySecretValidationError(
 				this.validator.errorsText(this.validateFunction.errors),
 			);
 		}
-		return parsedJson as RegistrySecrets;
```
pdcastro, line 84: Removing this return statement is a breaking change that requires `Change-type: major`, while this PR and commits are declared `Change-type: minor`. For example, the CLI relies on this return value:
https://github.com/balena-io/balena-cli/blob/v12.14.18/lib/utils/compose_ts.ts#L477

It's OK to have a breaking change and `Change-type: major`, but also it would be OK to simply leave the return statement there. :-)

lib/resolve.ts
```diff
@@ -104,3 +111,23 @@ export function resolveTask(
...
+		case 'i386':
+			return 'linux/386';
+		case 'aarch64':
+			return 'linux/arm64';
+		case 'armv7hf':
+			return 'linux/arm/v7';
+		case 'rpi':
+			return 'linux/arm/v6';
+	}
+	return null;
```
pdcastro, line 132: What about using an empty string instead of `null`? The "advantage" being that the return type is simpler as `string` instead of `string | null`. An empty string is a [falsy value](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), and could be tested with:
```
	if (platform) {
		task.dockerPlatform = platform;
	}
```
Or on a single line:
```
	task.dockerPlatform = platform || task.dockerPlatform;
```

pdcastro (changes requested):
* Does this PR provide the feature of "multiarch images" or "Docker image manifests"? Which would resolve CLI issues balena-io/balena-cli/issues/1408, balena-io/balena-cli/issues/1508 and possible others, and several support threads (paid and forum). I am going to test it with `balena build`...
* PR #81 got merged, so rebasing this PR on master may allow dropping one of the commits.

lib/index.ts
```diff
@@ -263,9 +263,13 @@ export async function performBuilds(
 		secrets: secretMap,
 		regSecrets: registrySecrets,
 		architecture,
+		experimental,
 	} = await initializeBuildMetadata(tasks, docker, tmpDir);
 
 	const images = await Bluebird.map(tasks, (task: BuildTask) => {
+		if (experimental) {
```
pdcastro, line 270: But what if an upcoming (or recently released?!) Docker version "promotes" the `platform` build option from experimental to standard? In fact, in Docker engine API v1.40, the `platform` option is not labeled experimental:
https://docs.docker.com/engine/api/v1.40/#operation/ImageBuild
(Unlike other options like "squash".)

If this happened, balena users' builds could suddenly start failing with "exec format error" simply because the updated Docker on their laptops to a newer version that "fully supports" the `platform` option in a non-experimental way.

What is the harm of adding the `platform` option to `task.dockerOpts`, if the engine is older and does not understand it? I would hope that the engine would simply ignore the option, but if it actually fails the build, then that's an argument for having the experimental test.

lib/resolve.ts
```diff
@@ -104,3 +111,23 @@ export function resolveTask(
...
+
+/**
+ * Given a balena architecture string, translate it to the equivalent
+ * docker platform string.
+ */
+function resolveDockerPlatform(balenaArchitecture: string): string | null {
+	switch (balenaArchitecture) {
+		case 'amd64':
+			return 'linux/amd64';
+		case 'i386':
```
pdcastro, line 123: On the following page, I see an additional balena architecture: `i386-nlp`
https://www.balena.io/docs/reference/base-images/devicetypes/

Should it be mapped to `linux/386` as well? Maybe @nghiant2710 knows the answer, as I've found issue balena-io/docs/issues/1165 :-)

lib/index.ts
```diff
@@ -263,9 +263,13 @@ export async function performBuilds(
...
+			task.dockerOpts = { platform: task.dockerPlatform, ...task.dockerOpts };
```
pdcastro, line 271: Something I found during CLI testing (`balena build` and `balena deploy --build` commands) is that the `pull: true` field is also needed in `dockerOpts` in order to cause Docker / balenaEngine to re-pull the image for a different platform / architecture, otherwise it will use any locally existing image that matches the image tag even if it is for the wrong architecture. This actually sounds like a Docker bug... I tested it with:
```
$ balena-engine version
Server:
 Engine:
  Version:          18.09.10-dev
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       7cb464a406748016f2df0c31a9851d20456a3d31
  Built:            Tue Mar 10 17:10:13 2020
  OS/Arch:          linux/amd64
  Experimental:     true
```
(Tested on Linux, where `docker` / balenaEngine does not use [binfmt_misc](https://en.wikipedia.org/wiki/Binfmt_misc).)

The `pull: true` option is described in:
* https://docs.docker.com/engine/api/v1.39/#operation/ImageBuild  
`pull  	Attempt to pull the image even if an older image exists locally.`
* https://docs.docker.com/engine/reference/commandline/build/#extended-description  
`docker build --pull`:  
`--pull  	Always attempt to pull a newer version of the image`

To put it another way, without the `pull: true` field in `dockerOpts`:
* `balena build test-multiarch --arch amd64 --deviceType intel-nuc --nocache`  
Assuming no matching base image existed before this command, a base image is correctly pulled for the `linux/amd64`. It works.
* `balena build test-multiarch --arch rpi --deviceType rpi --nocache`  
Running this command after the previous command, a base image already existed (for the `linux/amd64` platform) and therefore no new image is pulled for the `linux/arm/v6` platform. The `linux/amd64` image is used and the build fails with "exec format error". Issuing `docker rmi` command(s) for the base image(s), then re-running `balena build`, causes the `linux/arm/v6` image to be pulled and then `balena build` succeeds.

But I am also not saying that `pull: true` should be always there, as it could be a significant performance issue. I am considering adding a CLI option `balena build --pull` and warning messages advising users to add `pull` if the CLI host platform mismatches the application or device platform.

lib/registry-secrets.ts
robertgzr: not sure why this one was removed. will re-add it rather than break api

lib/resolve.ts
robertgzr: :ok_hand:

lib/index.ts
robertgzr: good point. we should be able to make that dependent on the api version as well I guess

lib/resolve.ts
robertgzr: I would love to have a definitive source on what platforms we support. probably should look through the supported device types to be sure

lib/index.ts
robertgzr: hmm that's messy. I'm not sure how to best deal with this... possibly the cli option + warnings is the best compromise. I also wouldn't want to be too smart and conditionally add the option

pdcastro: CLI PR balena-io/balena-cli/pull/2015 has added a `--pull` option to the `balena build` and `balena deploy` commands, so the CLI sets the `pull: true` Docker option when it contacts the Docker daemon directly. However, this does not cover `balena push`, where it would be for the builder to set to the `pull: true` Docker API option.

robertgzr: the reason for this would be the emulated builds right... I would actually like to be able to use `binfmt_misc` and specifying `balena build test-multiarch --arch amd64,aarch64`

robertgzr: on the other hand... did you try with a v19.03.x release of docker/balenaEngine?

pdcastro: I tested with Docker Desktop on macOS:
```
$ docker --version
Docker version 19.03.8, build afacb8b
```
But it never raises any `exec format error` because it always uses `binfmt_misc`, so I couldn't tell whether it was using the right base image or not. :-)   Would `uname -a` report the architecture of the image, or the architecture of the actual processor, when `binfmt_misc` is active?...

pdcastro: hmm I can think of some additional experiments I could try, like `balena deploy --build` with Docker 19.03, deploying to an RPi that should fail to execute the image if Docker selected an amd64 image...

pdcastro: I have answered some of my own questions:
* `uname -a` reports the architecture of the image, not the architecture of the actual processor when `binfmt_misc` is used for emulation, and therefore can be used to test the selection of base images.
* "Docker version 19.03.8, build afacb8b" (Docker Desktop for macOS) also requires `pull: true` (i.e. the `--pull` option in `balena build --pull` or `balena deploy --build --pull`) in order to refresh / select the correct base image architecture if a previous base image of same tag but different architecture already existed locally. (To me, this feels like a Docker bug, because Docker is effectively happy to use a base image for the wrong architecture.)

For `balena build` and `balena deploy`, I think we can get away with expecting users to add `--pull` in order for the correct base image to be downloaded, provided that a warning message is printed if the app / image architecture doesn't match the host architecture.

However, I am not yet clear about `balena push`, where `pull: true` (if needed) should be added by the balenaCloud builder.

lib/index.ts
```diff
@@ -293,14 +297,24 @@ export async function initializeBuildMetadata(
...
+			regSecrets: {},
+			architecture: '',
+			experimental: false,
+		};
 	}
 	// This feels a bit dirty, but there doesn't seem another
 	// nicer way to do it given the current setup
 	const buildMetadata = tasks[0].buildMetadata;
-	const architecture = (await docker.version()).Arch;
+	const versionOutput = await docker.version();
```
pdcastro, line 313: * The CLI already calls `docker.version()` before calling `performBuilds`, so will end up calling `docker.version()` twice. I wonder if it would make sense to take the result of `docker.version()` as an argument, instead of calling it here.
* In another comment, I question whether we should be testing for experimental. Testing the Docker API version might be more interesting. Maybe the Docker API version could be taken as argument, so there's no need to call `docker.version()` here.
* In that case, it might make more sense to get the API version as an argument to `performBuilds`, rather than an argument to `initializeBuildMetadata`, as the information is not used in this function.

lib/resolve.ts
```diff
@@ -104,3 +111,23 @@ export function resolveTask(
 
 	return task;
 }
+
+/**
+ * Given a balena architecture string, translate it to the equivalent
+ * docker platform string.
+ */
+function resolveDockerPlatform(balenaArchitecture: string): string | null {
```
pdcastro, line 119: It would be good to export this function, and further re-export it in `index.ts`:
```
export { resolveDockerPlatform } from './resolve';
```
The reason being that I found myself duplicating the logic in the CLI in order to conditionally print warning messages regarding the `--pull` and `--emulated` options when the app architecture doesn't match the docker daemon / balenaEngine host architecture.

lib/index.ts
kaisoz: @pdcastro the builder always does `pull: true` so we are safe in the cloud. But in the case of the cli, it feels wrong that the user has to pass always `--pull`. On the other hand, I wouldn't do an implicit `--pull` when building/deploying, since is an unexpected behavior

kaisoz: although to be honest, I'd like the builder to be consistent with the cli, will ease our lives from the maintenance point of view

lib/index.ts
kaisoz: @pdcastro Great idea, let's inject the docker version and use it as a threshold to use or not experimental features. The builder does `version` when generating deltas. Is it the same case for the cli? The builder could retrieve the version following the first connection to the builder, so in that way, it knows the worker capabilities since the beginning.

lib/index.ts
pdcastro: > But in the case of the cli, it feels wrong that the user has to pass always `--pull`. On the other hand, I wouldn't do an implicit `--pull` when building/deploying, since is an unexpected behavior.

Users don't have to pass always `--pull`, only when they want the effect of checking for base image updates / re-pulling images. :-) I read a comment from Page where he said that `--pull` does not necessarily cause a base image to be re-downloaded: it would only be re-downloaded if Docker detects that it has changed. And yet, testing with Docker Desktop on macOS, adding the `--pull` option causes a delay compatible with re-downloading. I need to double check whether or not a re-download was indeed happening.

This discussion re `--pull` behavior does not affect this `resin-multibuild` PR though. The `--pull` option was added to the CLI a few weeks ago (optional behavior), and "historically" (several years) the CLI does not pass the `pull` option to the Docker daemon. A change to this behavior would indeed be a breaking change.

lib/registry-secrets.ts
robertgzr: this slipped in here btw: https://github.com/balena-io-modules/resin-multibuild/pull/81/files#diff-af4b6a79d80562b9ccd9f13d18594741L84

lib/index.ts
```diff
@@ -316,7 +322,12 @@ export async function initializeBuildMetadata(
 		await populateSecrets(docker, secretMap, architecture, tmpDir);
 	}
 
-	return { secrets: secretMap, regSecrets: registrySecrets, architecture };
+	return {
+		secrets: secretMap,
+		regSecrets: registrySecrets,
+		architecture,
+		experimental,
```
pdcastro, line 329: `experimental` was left behind here

robertgzr: :sweat_smile:

test/build.spec.ts
```diff
@@ -226,14 +226,57 @@ describe('Resolved project building', () => {
...
+				expect(inspect)
+					.to.have.property('Architecture')
+					.that.equals('amd64');
+			});
+	});
+
+	it('should correctly build a resolved project for a different platform', function() {
+		if (
+			docker.version().then((version: any) => {
+				return version.experimental;
```
pdcastro, line 248: Is this missing some `await` clause? Because these lines would return immediately with a Promise object, with would the considered truthy and then the `if` test would always succeed. Also, the PR at the moment removed `experimental` detection, so this test should probably never be skipped. (?)

robertgzr: so I just prepend an `await`?

robertgzr:
```
		if (
			await docker.version().then((version: any) => {
				return version.experimental;
			})
		) {
			this.skip();
		}

```

robertgzr: I guess we can also just drop this check, no?

robertgzr: since we're not going to do that anyway...

lib/index.ts
```diff
@@ -266,6 +267,7 @@ export async function performBuilds(
 	} = await initializeBuildMetadata(tasks, docker, tmpDir);
 
 	const images = await Bluebird.map(tasks, (task: BuildTask) => {
+		task.dockerOpts = { platform: task.dockerPlatform, ...task.dockerOpts };
```
pdcastro, line 270: `task.dockerPlatform` could be `undefined`, so we could end up with:
```
    task.dockerOpts = { platform: undefined, ...task.dockerOpts };
```
So probably better code it as:
```
if (task.dockerPlatform) {
    task.dockerOpts = { platform: task.dockerPlatform, ...task.dockerOpts };
}
```

test/build.spec.ts
robertgzr: we can't drop it

robertgzr: the engine used by circleci is pretty old...
``
Server Engine Details:
  Version:          17.09.0-ce
  API version:      1.32 (minimum version 1.12)
  Go version:       go1.8.3
  Git commit:       afdb6d4
  Built:            2017-09-26T22:40:56.000000000+00:00
  OS/Arch:          linux/amd64
  Experimental:     false
``

lib/index.ts
pdcastro: @robertgzr, this `if (experimental)` test was removed as discussed, so that the image build API request always includes the `platform` option, despite the API documentation saying that it requires API version >= 1.34. We discussed that we should then test with Docker v17, v18, v19 on Windows / macOS / Linux, to see if the `balena build` command would result in error (from the Docker Engine). I have tested only on macOS, and got a somewhat unexpected result.

Up to Docker Desktop / Engine version 18.03.1 **(API version 1.37, Experimental: true),** the build failed with the following error:

```
$ docker version
Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:13:02 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.1-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   9ee9f40
  Built:        Thu Apr 26 07:22:38 2018
  OS/Arch:      linux/amd64
  Experimental: true

$ ./bin/balena-dev build ~/balena/other/simple_container/ -A armv7hf -d raspberrypi3 --pull --nocache
...
[Error]   Build failed.
(HTTP code 400) unexpected - invalid platform: invalid platform architecture "arm"
```

The next version of Docker Desktop for macOS available for download is 18.06.0 **(API version 1.38, Experimental: true),** which succeeds:

```
$ docker version
Client:
 Version:           18.06.0-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        0ffa825
 Built:             Wed Jul 18 19:05:26 2018
 OS/Arch:           darwin/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.0-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       0ffa825
  Built:            Wed Jul 18 19:13:46 2018
  OS/Arch:          linux/amd64
  Experimental:     true

$ ./bin/balena-dev build ~/balena/other/simple_container/ -A armv7hf -d raspberrypi3 --pull --nocache
...
[Build]   Built 1 service in 0:27
[Success] Build succeeded!
```

So my tentative conclusion is that we need to add a test, but not a test for Experimental nor for the API version (given that `Experimental: true` and API v1.37 fail to build), but rather a test for the Docker Engine version, which should be `>= 18.06.0`. Perhaps we should try testing on Linux as well, to further clarify this conclusion.

pdcastro: I have reproduced the results on Linux as well -- sort of, using docker-in-docker images. The build succeeds with Docker Engine v18.06.0 and later, and fails (`(HTTP code 400) unexpected - invalid platform: invalid platform architecture "arm"`) with v18.05.0 and earlier, even when `Experimental: false`.

**`balena build` fails with Docker v18.05.0 or earlier** (when the `platform` option is provided to the Docker Engine)
```
$ docker run --privileged --rm --name dind-18.05.0-ce -p 127.0.0.1:58992:2375 docker:18.05.0-ce-dind
...
time="2020-10-05T23:31:36.285042723Z" level=info msg="API listen on [::]:2375"

$ DOCKER_HOST=tcp://127.0.0.1:58992 docker version
...
Server:
 Engine:
  Version:          18.05.0-ce
  API version:      1.37 (minimum version 1.12)
  Go version:       go1.10.1
  Git commit:       f150324
  Built:            Wed May  9 22:20:42 2018
  OS/Arch:          linux/amd64
  Experimental:     false

$ DOCKER_HOST=tcp://127.0.0.1:58992 ./bin/balena-dev build ~/balena/other/simple_container/ -A armv7hf -d raspberrypi3 --pull --nocache
...
[Error]   Build failed.
(HTTP code 400) unexpected - invalid platform: invalid platform architecture "arm"
```

**`balena build` succeeds with Docker v18.06.0 or later** (when the `platform` option is provided to the Docker Engine)
```
$ docker run --privileged --rm --name dind-18.06.0-ce -p 127.0.0.1:58992:2375 docker:18.06.0-ce-dind
...
time="2020-10-05T23:32:21.139320655Z" level=info msg="API listen on [::]:2375"

$ DOCKER_HOST=tcp://127.0.0.1:58992 docker version
...
Server:
 Engine:
  Version:          18.06.0-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       0ffa825
  Built:            Wed Jul 18 19:13:39 2018
  OS/Arch:          linux/amd64
  Experimental:     false

 $ DOCKER_HOST=tcp://127.0.0.1:58992 ./bin/balena-dev build ~/balena/other/simple_container/ -A armv7hf -d raspberrypi3 --pull --nocache
...
[Build]   Built 1 service in 0:23
[Success] Build succeeded!
```

@robertgzr, @tmigone, it sounds to me like this PR should test for Docker Engine version 18.06.0 or later to decide whether or not to include the `platform` option in the docker options provided to the Docker Engine. I have not tested with balenaEngine, I assume it would be the same...

lib/index.ts
robertgzr: I created https://github.com/balena-os/balena-engine/issues/229 to track this possible bug in the engine

lib/index.ts
robertgzr: Note that we don't have any check on the current revision. It looks like you're testing on an amd64 host, for that case we will always need to use emulation, which is what we did before right?
If the user has `binfmt_misc` installed that won't be strictly necessary but I guess we would want to use qemu anyway....

pdcastro: > It looks like you're testing on an amd64 host, for that case we will always need to use emulation, which is what we did before right?

Yes. Adding the `--emulated` option to `balena build` makes no difference here though (I tested), because the error message:
```
(HTTP code 400) unexpected - invalid platform: invalid platform architecture "arm"
```
... appears to happen _before_ the Docker Engine starts building the image. The Docker Engine rejects the image build API request, as if during request validation: https://github.com/balena-os/balena-engine/blob/v18.04.0-ce-rc1/pkg/system/lcow.go#L15-L21

Indeed, HTTP code 400 means "Bad Request".

robertgzr: the thing is on balena engine 18.09.x and 19.03.x we should have the check for windows:
https://github.com/balena-os/balena-engine/blob/18.09-balena/pkg/system/lcow.go#L25
https://github.com/balena-os/balena-engine/blob/19.03-balena/pkg/system/lcow.go#L25

it does look like we need to check for version `18.06.0`, specifically because of this commit:
https://github.com/moby/moby/commit/7f334d3acfd7bfde900e16e393662587b9ff74a1

robertgzr: In that case let's restrict this to api version `>=1.38` which should get us out of trouble

robertgzr: I think with the latest commit we should be good

robertgzr:
I could use a hint in regards to:

```
lib/index.ts:309:4 - error TS2322: Type 'null' is not assignable to type 'SemVer'.


309  		apiVersion: null,

     		~~~~~~~~~~


  lib/index.ts:302:2

    302  apiVersion: semver.SemVer;

         ~~~~~~~~~~

    The expected type comes from property 'apiVersion' which is declared here on type '{ secrets: SecretsPopulationMap; regSecrets: RegistrySecrets; architecture: string; apiVersion: SemVer; }'


lib/index.ts:337:3 - error TS2322: Type 'SemVer | null' is not assignable to type 'SemVer'.

  Type 'null' is not assignable to type 'SemVer'.


337  	apiVersion,

     	~~~~~~~~~~


  lib/index.ts:302:2

    302  apiVersion: semver.SemVer;

         ~~~~~~~~~~

    The expected type comes from property 'apiVersion' which is declared here on type '{ secrets: SecretsPopulationMap; regSecrets: RegistrySecrets; architecture: string; apiVersion: SemVer; }'




```

lib/index.ts
```diff
@@ -297,18 +299,22 @@ export async function initializeBuildMetadata(
 	secrets: SecretsPopulationMap;
 	regSecrets: RegistrySecrets;
 	architecture: string;
+	apiVersion: semver.SemVer;
```
robertgzr, line 302: make this string? and coerce here: [`e869323` (#80)](https://github.com/balena-io-modules/resin-multibuild/pull/80/commits/e8693233c2b26202619c4830fa8391cbf3a2900e#diff-13b5b151431c7e7a17f273559ed212d5R272)

robertgzr: `semver.SemVer | null`

pdcastro (approved):

lib/index.ts
```diff
@@ -263,9 +265,17 @@ export async function performBuilds(
...
 		regSecrets: registrySecrets,
 		architecture,
+		apiVersion,
 	} = await initializeBuildMetadata(tasks, docker, tmpDir);
 
 	const images = await Bluebird.map(tasks, (task: BuildTask) => {
+		if (
+			task.dockerPlatform &&
+			apiVersion &&
+			semver.satisfies(apiVersion, '>=1.38.0')
```
pdcastro, line 275: 👍

PR #82, opened 2020-08-13
Revert breaking change to `validateRegistrySecrets` from 4.7.1

Page-:
Change-type: patch


Page-:
@balena-ci I self-certify!

PR #85, opened 2020-10-07
Add platform option in runBuildTask

robertgzr:
the previous PR #80 only added platform support to `performBuilds`

commit `Add platform in runBuildTask` (https://github.com/balena-io-modules/resin-multibuild/commit/d1f813b552f1587483cdbae12bab7a5c8e3a87dd) moves to the shared function ~~(which is also exported, so needs to be covered anyway)~~

~~I had to add another call to `docker.version()` in order to get the api version there... so I attempted to improve that by breaking API (as we discussed before)~~

~~https://github.com/balena-io-modules/resin-multibuild/commit/4cc18980506a175970788a960b578984dc044604 makes it the responsibility of the caller to fill `builderVersion` on a build task, which we then use here~~

Let me know what you think...

### DIFF
--- a/lib/build.ts
+++ b/lib/build.ts
@@ -2,6 +2,7 @@ import * as Promise from 'bluebird';
 import * as Dockerode from 'dockerode';
 import * as _ from 'lodash';
 import { Builder, BuildHooks, FromTagInfo } from 'resin-docker-build';
+import * as semver from 'semver';
 import * as Stream from 'stream';
 
 import { SecretsPopulationMap } from './build-secrets';
@@ -90,20 +91,36 @@ const generateLabels = (task: BuildTask): { labels?: Dictionary<string> } => {
  * @param docker The handle to the docker daemon
  * @return a promise which resolves to a LocalImage which points to the produced image
  */
-export function runBuildTask(
+export async function runBuildTask(
 	task: BuildTask,
 	docker: Dockerode,
 	registrySecrets: RegistrySecrets,
 	secrets?: SecretsPopulationMap,
 	buildArgs?: Dictionary<string>,
 ): Promise<LocalImage> {
-	// First merge in the registry secrets (optionally being
-	// overridden by user input) so that they're available for
-	// both pull and build
-	task.dockerOpts = _.merge(
-		{ registryconfig: registrySecrets },
-		task.dockerOpts,
-	);
+	// check if docker supports propagating which platform to build for
+	// NOTE: docker api version 1.34 actually introduced platform to the
+	// api but it was broken until fixed in
+	// https://github.com/moby/moby/commit/7f334d3acfd7bfde900e16e393662587b9ff74a1
+	// which is why we check for 1.38 here
+	const usePlatformOption: boolean =
+		!!task.dockerPlatform &&
+		semver.satisfies(
+			semver.coerce((await docker.version()).ApiVersion) || '0.0.0',
+			'>=1.38.0',
+		);
+
+	task.dockerOpts = {
+		// First merge in the registry secrets (optionally being
+		// overridden by user input) so that they're available for
+		// both pull and build
+		registryconfig: registrySecrets,
+		// then merge in the target platform to ensure pullExternal
+		// also considers it
+		...(usePlatformOption ? { platform: task.dockerPlatform } : {}),
+		...task.dockerOpts,
+	};
+
 	if (task.external) {
 		// Handle this separately
 		return pullExternal(task, docker);
--- a/lib/build.ts
+++ b/lib/build.ts
@@ -1,4 +1,3 @@
-import * as Promise from 'bluebird';
 import * as Dockerode from 'dockerode';
 import * as _ from 'lodash';
 import { Builder, BuildHooks, FromTagInfo } from 'resin-docker-build';
--- a/test/build.spec.ts
+++ b/test/build.spec.ts
@@ -292,13 +292,13 @@ describe('Invalid build input', () => {
 			.then(() => {
 				throw new Error('Error not thrown on null buildStream input');
 			})
-			.catch(BuildProcessError, () => {
-				// This is what we want
-			})
 			.catch(e => {
-				throw new Error(
-					'Incorrect error thrown on null buildStream input: ' + e,
-				);
+				// This is what we want
+				if (!(e instanceof BuildProcessError)) {
+					throw new Error(
+						'Incorrect error thrown on null buildStream input: ' + e,
+					);
+				}
 			});
 	});
 });
--- a/test/build.spec.ts
+++ b/test/build.spec.ts
@@ -258,6 +258,7 @@ describe('Resolved project building', () => {
 			serviceName: 'test',
 			streamHook: streamPrinter,
 			buildMetadata,
+			dockerOpts: { pull: true },
 		};
 		return new Promise((resolve, reject) => {
 			const resolveListeners = {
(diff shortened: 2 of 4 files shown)


lib/build.ts
```diff
@@ -104,6 +105,20 @@ export function runBuildTask(
...
+	// we build against a docker daemon that supports it.
+	if (
+		task.dockerPlatform &&
+		task.builderVersion &&
+		semver.satisfies(
+			semver.coerce(task.builderVersion.ApiVersion) ?? '0.0.0',
+			'>=1.38.0',
+		)
+	) {
+		task.dockerOpts = { platform: task.dockerPlatform, ...task.dockerOpts };
```
pdcastro, line 119: There is already a `task.dockerOpts = _.merge(...)` call above, which by the way unnecessarily uses `lodash`. Suggestion:
```
	const usePlatform: boolean =
		task.dockerPlatform &&
		task.builderVersion &&
		semver.satisfies(
			semver.coerce(task.builderVersion.ApiVersion) ?? '0.0.0',
			'>=1.38.0',
		);

	task.dockerOpts = {
		registryconfig: registrySecrets,
		...(usePlatform ? { platform: task.dockerPlatform } : {}),
		...task.dockerOpts,
	};
```

pdcastro (changes requested):

lib/build.ts
```diff
@@ -90,20 +91,37 @@ const generateLabels = (task: BuildTask): { labels?: Dictionary<string> } => {
...
  * @return a promise which resolves to a LocalImage which points to the produced image
  */
-export function runBuildTask(
+export async function runBuildTask(
 	task: BuildTask,
 	docker: Dockerode,
 	registrySecrets: RegistrySecrets,
 	secrets?: SecretsPopulationMap,
 	buildArgs?: Dictionary<string>,
 ): Promise<LocalImage> {
```
robertgzr, line 100:
```
Duplicate identifier 'Promise'. Compiler reserves name 'Promise' in top level scope of a module containing async functions.

1 import * as Promise from 'bluebird';
              ~~~~~~~

lib/build.ts:100:4 - error TS1064: The return type of an async function or method must be the global Promise<T> type.

100 ): Promise<LocalImage> {
       ~~~~~~~~~~~~~~~~~~~
```

@pdcastro looks lik it didn't like that I just made the function async

pdcastro: Ah, sorry, hadn't realised it was using Bluebird promises instead of native promises - at the top of the file:
```
import * as Promise from 'bluebird';
```
So, either use `.then()` instead of async/await or, probably best, delete the line above (`import * as Promise from 'bluebird';`) so that `Promise` refers to a native promise instead of a Bluebird promise. Up to you!

pdcastro: It doesn't look like any Bluebird features are being used - neither inside `runBuildTask` nor by its callers - so it should be safe to delete the `import * as Promise from 'bluebird'` line at the top of the file (which would then allow using `async/await`).

test/build.spec.ts
```diff
@@ -292,8 +292,9 @@ describe('Invalid build input', () => {
 			.then(() => {
 				throw new Error('Error not thrown on null buildStream input');
 			})
-			.catch(BuildProcessError, () => {
+			.catch((e: any): e is BuildProcessError => {
```
pdcastro, line 295: hmm I don't think this achieves the same... As coded above, I think it will catch any errors of any type, and will cast the error (of any type) to `BuildProcessError` -- i.e. simply disabling type checking.

I believe that the change that would preserve the Bluebird behavior of only catching `BuildProcessError` would be to replace the two catch blocks with a single one as follows:

``` 
	.catch((e: Error) => {
		if (!(e instanceof BuildProcessError)) {
			throw new Error(
				'Incorrect error thrown on null buildStream input: ' + e,
			);
		}
	});
```

pdcastro (changes requested):

lib/build.ts
```diff
@@ -90,20 +90,36 @@ const generateLabels = (task: BuildTask): { labels?: Dictionary<string> } => {
...
+	// check if docker supports propagating which platform to build for
+	// NOTE: docker api version 1.34 actually introduced platform to the
+	// api but it was broken until fixed in
+	// https://github.com/moby/moby/commit/7f334d3acfd7bfde900e16e393662587b9ff74a1
+	// which is why we check for 1.38 here
+	const usePlatformOption: boolean =
+		task.dockerPlatform &&
+		semver.satisfies(
+			semver.coerce((await docker.version()).ApiVersion),
+			'>=1.38.0',
```
pdcastro, line 109: Build is failing here:
```
lib/build.ts:105:8 - error TS2322: Type 'boolean | "" | undefined' is not assignable to type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.

105  const usePlatformOption: boolean =
           ~~~~~~~~~~~~~~~~~

lib/build.ts:108:4 - error TS2345: Argument of type 'SemVer | null' is not assignable to parameter of type 'string | SemVer'.
  Type 'null' is not assignable to type 'string | SemVer'.

108  		semver.coerce((await docker.version()).ApiVersion),
     		~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I think this solves it:
```
	const usePlatformOption: boolean =
		!!task.dockerPlatform &&
		semver.satisfies(
			semver.coerce((await docker.version()).ApiVersion) || '0.0.0',
			'>=1.38.0',
		);
```
(Note I also had to add `!!` in front of `task.dockerPlatform`.)

test/build.spec.ts
pdcastro: > I don't think this achieves the same...

Confirmed by looking at the compiled code (Typescript to Javascript, `npm run buld:test`, looking at `build/test/build.spec.js`). The following is not good: :-)

```
        return build_1.runBuildTask(task, docker, secretMap, buildVars)
            .then(() => {
            throw new Error('Error not thrown on null buildStream input');
        })
            .catch((e) => {
            return;
        })
            .catch(e => {
            throw new Error('Incorrect error thrown on null buildStream input: ' + e);
        });
```

robertgzr: :facepalm:

lib/build.ts
robertgzr: thanks!

test/build.spec.ts
```diff
@@ -292,13 +292,13 @@ describe('Invalid build input', () => {
...
-				);
+				// This is what we want
+				if (!(e instanceof BuildProcessError)) {
+					throw new Error(
+						'Incorrect error thrown on null buildStream input: ' + e,
+					);
+				}
 			});
 	});
 });
```
pdcastro, line 304: Although the tests are passing by the CI, they fail on my laptop with Docker Desktop for macOS:
```
  1) Resolved project building
       should correctly build a resolved project for a different platform:

      AssertionError: expected 'amd64' to equal '386'
      + expected - actual

      -amd64
      +386

      at /Users/paulo/balena/balena-io-modules/resin-multibuild/build/test/build.spec.js:241:27
```
```
$ docker version
...
Server: Docker Engine - Community
 Engine:
  Version:          19.03.13
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       4484c46d9d
  Built:            Wed Sep 16 17:07:04 2020
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v1.3.7
  GitCommit:        8fba4e9a7d01810a393d5d25a3621dc101981175
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```
Trying to figure out why...

kaisoz (approved):

pdcastro (approved):

PR #86, opened 2020-10-18
Avoid overwriting docker build options with undefined values

pdcastro:
This PR fixes a "potential" bug. I have not actually come across a problem in practice, but while integrating the latest version of multibuild in the balena CLI, I noticed that the `platform` and `registryconfig` docker build options could be overwritten with `undefined` (as explained in a comment) which is not ideal and didn't happen before PR #85.

Change-type: patch

### DIFF
--- a/lib/build.ts
+++ b/lib/build.ts
@@ -109,16 +109,16 @@ export async function runBuildTask(
 			'>=1.38.0',
 		);
 
-	task.dockerOpts = {
+	task.dockerOpts = _.merge(
 		// First merge in the registry secrets (optionally being
 		// overridden by user input) so that they're available for
 		// both pull and build
-		registryconfig: registrySecrets,
+		{ registryconfig: registrySecrets },
 		// then merge in the target platform to ensure pullExternal
 		// also considers it
-		...(usePlatformOption ? { platform: task.dockerPlatform } : {}),
-		...task.dockerOpts,
-	};
+		usePlatformOption ? { platform: task.dockerPlatform } : {},
+		task.dockerOpts,
+	);
 
 	if (task.external) {
 		// Handle this separately
--- a/test/build.spec.ts
+++ b/test/build.spec.ts
@@ -265,6 +265,11 @@ describe('Resolved project building', () => {
 				error: [reject],
 			};
 			const newTask = resolveTask(task, 'i386', 'qemux86', resolveListeners);
+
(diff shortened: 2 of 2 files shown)


lib/build.ts
```diff
@@ -109,16 +109,16 @@ export async function runBuildTask(
 			'>=1.38.0',
 		);
 
-	task.dockerOpts = {
+	task.dockerOpts = _.merge(
```
pdcastro, line 112: In PR #85, I incorrectly advised that lodash's `_.merge` could be replaced with the `...` spread syntax, but there are corner cases:
* If `task.dockerOpts` contains properties set to `undefined`, `_.merge` skips them while the spread syntax sets the option to `undefined`. For example:
```
> _.merge({ a: 1 }, { a: undefined }, undefined, { b: 2 })
{ a: 1, b: 2 }

> { a: 1, ...{ a: undefined } }
{ a: undefined }
```

* `_.merge` recursively merges objects in depth, so that `registryconfig` values would be added together rather than replaced:

```
> _.merge({ o: { b: 2 } }, { o: { c: 3 } })
{ o: { b: 2, c: 3 } }

> {o: { b: 2 }, ...{ o: { c: 3 } } }
{ o: { c: 3 } }
```

kaisoz: Good catch! could we add a test so that we are sure that this doesn't happen again in a future change? 

There's a trend towards dropping the use of lodash functions in Balena... and given how we work, it could happen that some future developer replaces this back to the spread syntax. A test could prevent this :+1:

pdcastro: Thanks. I've modified an existing test in `build.spec.ts` to cover the case of `task.dockerOpts` containing a `platform: undefined` value.

test/build.spec.ts
```diff
@@ -265,6 +265,11 @@ describe('Resolved project building', () => {
 				error: [reject],
 			};
 			const newTask = resolveTask(task, 'i386', 'qemux86', resolveListeners);
+
+			// also test that a `platform: undefined` value in `task.dockerOpts`
+			// does not override a valid value in `task.dockerPlatform`
+			task.dockerOpts.platform = undefined;
```
pdcastro, line 271: With this modification to the test, the test would fail without this PR's code changes to `lib/build.ts`.

robertgzr (approved):
good catch

PR #88, opened 2021-04-30
Conditionally bypass Docker 'platform' option, update dependencies

pdcastro:
Two commits (in the same PR because they were tested together) -
* One commit allows the Docker build image API `platform` option to be bypassed by setting `task.dockerPlatform` to `'none'`, which `balena-cli` can do to avoid `balena build` failing when used with Docker v20.10. Related to: https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc
This was necessary because, otherwise, the CLI could not use the latest version of `balena-multibuild`, including the other commit in this PR.

* Bump `docker-progress` and remove the dependency on `docker-toolbelt`, as explained in PR comments and as part of the investigation of issue balena-io/balena-cli/issues/2165 (Node.js v14 support by the balena CLI).

### DIFF
--- a/lib/build.ts
+++ b/lib/build.ts
@@ -121,6 +121,8 @@ export async function runBuildTask(
 	// which is why we check for 1.38 here
 	const usePlatformOption: boolean =
 		!!task.dockerPlatform &&
+		// workaround for `balena build` with Docker v20.10
+		task.dockerPlatform !== 'none' &&
 		semver.satisfies(
 			semver.coerce((await docker.version()).ApiVersion) || '0.0.0',
 			'>=1.38.0',
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -67,3 +67,44 @@ export function generateBuildTasks(
 		}
 	});
 }
+
+/**
+ * Separate string containing registry and image name into its parts.
+ * Example:
+ *   getRegistryAndName('registry.balena-staging.com/resin/rpi')
+ *   -> { registry: "registry.balena-staging.com", imageName: "resin/rpi" }
+ *
+ * This function was copied here from the `docker-toolbelt` module because it
+ * was the only reason for `resin-multibuild` and the balena CLI to import
+ * that module. TODO: modernize the `docker-toolbelt` module, including
+ * deleting legacy code and converting it to TypeScript.
+ */
+export function getRegistryAndName(image: string) {
+	// Matches (registry)/(repo)(optional :tag or @digest)
+	// regex adapted from Docker's source code:
+	// https://github.com/docker/distribution/blob/release/2.7/reference/normalize.go#L62
+	// https://github.com/docker/distribution/blob/release/2.7/reference/regexp.go#L44
+
+	const match = image.match(
+		/^(?:(localhost|.*?[.:].*?)\/)?(.+?)(?::(.*?))?(?:@(.*?))?$/,
+	);
+	if (!match) {
+		throw new Error(`Could not parse image "${image}"`);
+	}
+	const [, registry, imageName, tag, digest] = match;
+	const tagName = !digest && !tag ? 'latest' : tag;
+	const digestMatch = digest?.match(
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -52,8 +52,7 @@
   "dependencies": {
     "ajv": "^6.12.3",
     "bluebird": "^3.7.2",
-    "docker-progress": "^4.0.1",
-    "docker-toolbelt": "^3.3.8",
```
pdcastro, line 56: `docker-toolbelt` declares its own dependency on dockerode v2.5.8, but I wanted to bump `dockerode` to v3 in the balena CLI -- and actually make use of dockerode v3 through `balena-multibuild`, rather than just have multiple copies of different versions of `dockerode` under the `node_modules` folder. I wanted to do this as part of the investigation on Node.js v14 support -- issue balena-io/balena-cli/issues/2165 -- and in any case, I expect that sooner or later we will have to update to dockerode v3. Updating `docker-toolbelt` would be an option, but it is in need of modernization (some legacy code can be deleted, coffeescript -> typescript conversion, bump dependencies) while I realized that the only reason to import it here and in `balena-cli` was to make use of the `getRegistryAndName` function, a synchronous regex utility function that does not have any dependencies of its own. So I copied that small-ish function to `balena-multibuild`, exported it for use by `balena-cli`, and removed the dependency on `docker-toolbelt` altogether.

I have only partially achieved the goal of using dockerode v3 through dependency injection in the CLI and `balena-multibuild`: both `docker-progress` and `balena-multibuild` now use dependency injection for the `dockerode` instance, so that the CLI can inject `dockerode` v3. However, `balena-multibuild` still declares a dependency on `dockerode` v2.5.8 for compatibility with [resin-docker-build](https://github.com/balena-io-modules/resin-docker-build). The next step would be to update `resin-docker-build`, but I need to work on other higher priority tasks and cannot continue this investigation right now. So I decided to merge the improvements I have made so far (this PR and a few others).

lib/utils.ts
```diff
@@ -67,3 +67,44 @@ export function generateBuildTasks(
...
+ * Example:
+ *   getRegistryAndName('registry.balena-staging.com/resin/rpi')
+ *   -> { registry: "registry.balena-staging.com", imageName: "resin/rpi" }
+ *
+ * This function was copied here from the `docker-toolbelt` module because it
+ * was the only reason for `resin-multibuild` and the balena CLI to import
+ * that module. TODO: modernize the `docker-toolbelt` module, including
+ * deleting legacy code and converting it to TypeScript.
+ */
+export function getRegistryAndName(image: string) {
```
pdcastro, line 82: See jsdoc comments above, and the PR comment for `package.json`.

package.json
pdcastro, line 56: This PR removes the dependency on `docker-toolbelt`.

lib/build.ts
```diff
@@ -121,6 +121,8 @@ export async function runBuildTask(
 	// which is why we check for 1.38 here
 	const usePlatformOption: boolean =
 		!!task.dockerPlatform &&
+		// workaround for `balena build` with Docker v20.10
+		task.dockerPlatform !== 'none' &&
```
pdcastro, line 125: @robertgzr @dfunckt @toochevere, you may remember or be aware that last year we added the ability of selecting multiarch base images to `resin-multibuild` and this got deployed to `balena-builder` and became usable through `balena push`. But just before we got the CLI / `balena build` to use it, we found that it would break `balena build` when used with Docker v20.10. That very long thread: https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc

This change (and similar in `resolve.ts`) allows the CLI to bypass the "multiarch feature" (and keep `balena build` working without support for multiarch base images) if the CLI explicitly sets `task.dockerPlatform` to the string `'none'`, which I am doing in this other CLI PR:
https://github.com/balena-io/balena-cli/pull/2259

robertgzr (approved):
lgtm for the platform passing, I'll leave the rest for somebody more knowledgeable in TS :)

robertgzr:
I guess that's that ^^

PR #92, opened 2021-08-05
Determine when to use the docker 'platform' flag based on image manifest

toochevere:
This change is part of the larger effort to enable mult-arch image support.

Change-type: minor
Signed-off-by: Paul Jonathan <pj@balena-io>


lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+//            close it's eyes and hope for the best.  This opens the possibility for an exec
+//            error, but also allows users to continue to use v1 images if they want to / need to.
+//
+async function getRecommendedPlatformhHandling(
+	task: BuildTask,
+	docker: Dockerode,
+): Promise<'allowEnginePlatformHandling' | 'noPlatformHandling' | 'unknown'> {
+	let imageReferences: string[];
+
+	const getHasV1Manfiest = async (repository: string) => {
```
dfunckt, line 261: `getHas...` is a bit awkward, Java/C# is strong in you padawan :) There's also a typo.

```suggestion
	const hasV1Manifest = async (repository: string) => {
```

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+		const registry = new RegistryClient(docker.modem);
+		try {
+			const manifest = await registry.getManifest(repository);
+			return (
+				(manifest as DockerImageManifest).Descriptor.mediaType ===
```
dfunckt, line 266: You shouldn't need to coerce `manifest`. Why is that?

lib/registry/registry-client.ts
```diff
@@ -0,0 +1,72 @@
+/**
```
dfunckt, line 1: This file is really only about image manifests, I feel `manifests.ts` is a more appropriate file name.

lib/index.ts
```diff
@@ -60,6 +60,7 @@ export { getRegistryAndName } from './utils';
 export { BalenaYml, ParsedBalenaYml };
 export { PathUtils };
 export { ResolveListeners };
+export { CANONICAL_HUB_URL } from './constants';
```
dfunckt, line 63: What's the need to make this constant public API?

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+//   This method attempts to calculate if we should pass the `platform` flag to Docker, or
+//   not.
+//
+///  The complexity arises from the fact that old v1 manifests do not describe architecture.
+//   In such csaes, Docker will try to guess based on the platform of the image based
+//   on the current machine arch.
+//
+//   When Docker recieves a target platform, these are the possible cases:
+//
+//      + V2 manifest for all images, and all images support the requested arch
```
robertgzr, line 213: I would make it clear that we require a manifest list/image index here

robertgzr (changes requested):
nice. I just added a bunch of comments mostly around the big comment documenting all of this. btw do you think it might be helpful to extract that into a markdown file? I don't know if this info is accessible in this condensed form anywhere else...

lib/build.ts
```diff
@@ -20,13 +20,19 @@ import * as _ from 'lodash';
 import { Builder, BuildHooks, FromTagInfo } from 'resin-docker-build';
 import * as semver from 'semver';
 import type * as Stream from 'stream';
+import { DockerfileParser } from 'dockerfile-ast';
```
pdcastro, line 23: `DockerfileParser` seems to be used only inside function `getRecommendedPlatformhHandling`, and even then only in a certain decision branch, so I suggest moving the import clause to the line immediately before the point where it is used:
```
    const { DockerfileParser } = await import('dockerfile-ast')
    const parsedDockerfile = DockerfileParser.parse(task.dockerfile);
```

You will find inconsistencies in the codebase, but in general we prefer to do this because it improves "responsiveness" (performance -- time between typing the command and getting some response), especially when there are code execution paths that might no need the module. For example, `if (task.imageName)` is true, then there is no need for the CLI to spend time parsing and importing `'dockerfile-ast'`, as it won't be used. This is generally not an issue in long-running web applications (like balena-builder), but it's an issue in interactive applications like the balena CLI. While a single import clause doesn't mean much, it adds up when this approach is followed throughout the codebase.

lib/build.ts
```diff
@@ -20,13 +20,19 @@ import * as _ from 'lodash';
...
 import type { BuildTask } from './build-task';
 import { BuildProcessError } from './errors';
 import { pullExternal } from './external';
 import { LocalImage } from './local-image';
 import type { RegistrySecrets } from './registry-secrets';
+import {
+	DockerImageManifest,
+	RegistryClient,
+	MEDIATYPE_MANIFEST_V1,
+} from './registry/registry-client';
```
pdcastro, line 35: Similar to my other comment, this one could be split in two:

* If needed to annotate function signatures, an `import type` clause here that is only used by TypeScript for type checking and disappears in the compiled JavaScript code, thus having no runtime implications:
```ts
import type { DockerImageManifest } from './registry/registry-client';
```
* An import clause inside `getRecommendedPlatformhHandling`, closer to where it gets used:
```
const { RegistryClient, MEDIATYPE_MANIFEST_V1 } = await import('./registry/registry-client');
```

lib/registry/registry-client.ts
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import * as Dockermodem from 'docker-modem';
```
pdcastro, line 19: NIT: Would `import type` (rather than just `import`) work here? The only advantage is that `import type` makes it clear to the reader that this is only for TypeScript type checking and will disappear in compiled JavaScript.

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+
+	if (!hasAnyV1Manifest) {
+		return 'allowEnginePlatformHandling';
+	}
+	if (hasBalenaImages) {
+		throw new BuildProcessError(
+			"Using outdated 'balenalib' images that do not support platform targetting.  Please update to the latest `balenalib` images",
+		);
+	}
+	return 'noPlatformHandling';
```
pdcastro, line 323: So, in the case of mixed "bad manifest" (hasAnyV1Manifest is true) and "good manifest" multiarch images (say, `debian`), where the "bad manifest" images are **not** balenalib images, this function would return `'noPlatformHandling'`. This could result in the incorrect image architecture being selected for the good manifest images (e.g. Intel instead of ARM), and "exec format error" when the app is deployed to the fleet. In this case, I think we should print a warning message to help the user figure things out:
```
Warning: The following images do not support architecture selection:
- Service "myService1", Image "badrepo1/badimage1"
- Service "myService1", Image "badrepo2/badimage2"

While the following images require architecture selection:
- Service "MyService1", Image, "goodrepo1/goodimage1"
- Service "MyService1", Image, "goodrepo2/goodimage2"

To avoid an image build error, the CLI has disabled architecture selection.
However, this means that the architecture of the machine where the Docker
Engine is running will be used by default to select the architecture of the
images that require architecture selection, which may result in "exec format
error" at runtime. To avoid this problem, please update or replace the images
that do not support architecture selection.
```

pdcastro (commented):
Suggestion: run a spellchecker on the comments, as I found a few typos:
recieves -> receives
assumee -> assume
bests -> best
it's eyes -> its eyes

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+//          - If we have balenalib images, we know that they have been updated.  Throw an error
+//            and tell the user to update
+//          - If we do not have balenalib images, don't pass the platform flag.  Let Docker
+//            close it's eyes and hope for the best.  This opens the possibility for an exec
+//            error, but also allows users to continue to use v1 images if they want to / need to.
+//
+async function getRecommendedPlatformhHandling(
+	task: BuildTask,
+	docker: Dockerode,
+): Promise<'allowEnginePlatformHandling' | 'noPlatformHandling' | 'unknown'> {
```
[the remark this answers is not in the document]
toochevere: It could, but I felt like using enumerated values was more self-documenting.  Let me know if you feel strongly about using `boolean`

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+	let imageReferences: string[];
+
+	const getHasV1Manfiest = async (repository: string) => {
+		const registry = new RegistryClient(docker.modem);
+		try {
+			const manifest = await registry.getManifest(repository);
```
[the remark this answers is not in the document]
toochevere: I was debating about this.  I'll go with what you suggest.

lib/index.ts
toochevere: This ended up as part of some refactoring that sort of took place as I was going.  It is used in the builder.

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+//
+//      + V2 manifest for all images, and all images support the requested arch
+//        Note that "support" means either it is a single arch image with the correct arch
+//        or else it is a multi-arch image with one of the images matching arch
+//          -> Result: Good to go
+//
+//      + V2 manifest for all images, and some do not suppor the requested arch
+//          -> Result: Docker error
+//
+//      + Some V1 manifests, the plaform flag matches what Docker guesses
```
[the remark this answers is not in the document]
toochevere: I don't.  I'll have to find it.  Good idea.

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
...
+	}
+
+	let hasBalenaImages = false;
+	let hasAnyV1Manifest = false;
+
+	await Promise.all(
+		imageReferences.map(async (r) => {
+			const thisImageHasV1Manifest = await getHasV1Manfiest(r);
+			hasAnyV1Manifest = thisImageHasV1Manifest || hasAnyV1Manifest;
+			hasBalenaImages = r.startsWith('balenalib/') || hasBalenaImages;
```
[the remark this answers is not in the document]
toochevere: good catch

lib/build.ts
```diff
@@ -187,3 +200,125 @@ export async function runBuildTask(
 		});
 	});
 }
+
+//   This method attempts to calculate if we should pass the `platform` flag to Docker, or
+//   not.
+//
+///  The complexity arises from the fact that old v1 manifests do not describe architecture.
```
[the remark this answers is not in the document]
toochevere: @robertgzr it's kind of interesting.  Even though the media type is the same, the format that comes back when going through the docker daemon is a bit different from what I was getting when querying the registry directly.   You're right in that now I do see platform coming through, but it's potentially wrong.

As to how often we see v1 coming through, I can't say definitively.  That question kind of goes to the heart of why we are even doing this.  If we felt it was extremely rare, could we completely cut out asking the registry?  I'm inclined to say we should avoid potentially breaking images that aren't up to date.

lib/build.ts
toochevere: @pdcastro this then result in the same code getting called each time `getRecommendedPlatformHandling` (once per call of `runBuildTask`.  Seems that, although the point is well taken especially in the previous case, perhaps it might be non-performant here?

pdcastro: I think you are right, mainly because of how `balena-builder` uses `balena-multibuild`. It is a trade off between spending import time when `balena-multibuild` is imported for the first time (which contributes to interactive CLI usage "feeling slow"), _vs._ import time when `getRecommendedPlatformHandling` is called (which is worse for long-lived processes like `balena-builder`).  The result of `import('./registry/registry-client')` will be cached, so only the first `import` call will be relatively expensive, but even so it costs something. For `balena-cli`, `getRecommendedPlatformHandling` would be called a small-ish number of time, say 20 times for 20 services (devenv / BoB), and each call would dilute the import time with the image build time for the service that can take minutes. But for `balena-builder`, `getRecommendedPlatformHandling` could be called thousands of times over weeks, months, years...

lib/build.ts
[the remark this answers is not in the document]
toochevere: @robertgzr I tried finding this in the Moby source and it was a bit of a rabbit hole.  I think I found what might be the place where this is happening, but it appears to just take whatever the registry hands back, which would mean that the registry is making assumptions?   Sorry, this is my first foray into Go and the Moby project is definitely the deep end of the pool.
https://github.com/moby/moby/blob/7a50fe8a52f4e6fc6a1e6624ca81defc1e69d748/api/server/router/distribution/distribution_routes.go#L143

lib/build.ts
toochevere: Ok, good idea.  But the downside is that we lose any short-cutting; gotta check each image now.

lib/build.ts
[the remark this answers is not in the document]
toochevere: Renaming this method for clarity and going with boolean as suggested.

lib/build.ts
toochevere: Ah, ok.  Didn't know that about the caching.  Good to know.

lib/build.ts
```diff
@@ -187,3 +194,169 @@ export async function runBuildTask(
...
+			'Error: The following balenalib images do not support platform targeting. Please update them:\n-' +
+				outdatedBalenalibImages.join('\n-'),
+		);
+	}
+	if (v1Images.length === 0) {
+		// Only v2 images, let Docker receive `--platform`
+		return true;
+	} else if (v2Images.length > 0) {
+		// Mixed v1 and v2 images
+		console.warn(
```
toochevere, line 331: Is this the proper way to emit a warning?

lib/build.ts
```diff
@@ -187,3 +194,169 @@ export async function runBuildTask(
...
+		}
+		imageReferences = [];
+		fromInstructions.forEach((inst) => {
+			const fromLineArguments = inst.getArguments();
+			if (fromLineArguments.length === 0) {
+				// bail out, no image specified?
+				return;
+			}
+
+			imageReferences.push(fromLineArguments[0].getValue());
```
klutchell, line 297: Do we really have to specify the index here? I think this will break if additional args are provided in the FROM line.
```
FROM --platform=arm/v7 alpine AS build
```
I think the above Dockerfile will result in us passing `--platform=arm/v7` to the list of images, but I could be mistaken.

klutchell: https://docs.docker.com/engine/reference/builder/#from

lib/build.ts
```diff
@@ -187,3 +194,169 @@ export async function runBuildTask(
...
+				v1Images.push(r);
+			} else {
+				v2Images.push(r);
+			}
+		}),
+	);
+
+	if (outdatedBalenalibImages.length > 0) {
+		throw new BuildProcessError(
+			'Error: The following balenalib images do not support platform targeting. Please update them:\n-' +
```
klutchell, line 322: I think a warning here is more realistic than forcing folks to update their pinned versions. We don't want to discourage version pinning in any way. What if the latest images break something for them?

lib/build.ts
```diff
@@ -187,3 +194,169 @@ export async function runBuildTask(
...
+				v1Images.join('\n-') +
+				'\n\n' +
+				'The following images require architecture selection:\n-' +
+				v2Images.join('\n-') +
+				'\n\n' +
+				'To avoid an image build error, the CLI has disabled architecture selection. ' +
+				'However, this means that the architecture of the machine where the Docker ' +
+				'Engine is running will be used by default to select the architecture of the ' +
+				'images that require architecture selection, which may result in "exec format ' +
+				'error" at runtime. To avoid this problem, please update or replace the images ' +
```
klutchell, line 343:
```suggestion
				'error" at runtime. To avoid this warning, please update or replace the images ' +
```
It's more likely they want the warning to go away, as they might not consider it to be a "problem" even if we do.

lib/build.ts
```diff
@@ -187,3 +194,169 @@ export async function runBuildTask(
...
+		// Mixed v1 and v2 images
+		console.warn(
+			'Warning: Dockerfile contains a mixes images that require an architecture selection with ones that do not support it.\n' +
+				'The following images do not support architecture selection:\n-' +
+				v1Images.join('\n-') +
+				'\n\n' +
+				'The following images require architecture selection:\n-' +
+				v2Images.join('\n-') +
+				'\n\n' +
+				'To avoid an image build error, the CLI has disabled architecture selection. ' +
```
klutchell, line 339:
```suggestion
				'To avoid an image build error, the CLI has disabled platform selection. ' +
```

lib/build.ts
```diff
@@ -187,3 +194,169 @@ export async function runBuildTask(
...
+	} else {
+		// Only old v1 images
+		console.warn(
+			'Warning: Dockerfile contains a images do not support architecture selection.\n-' +
+				v1Images.join('\n-') +
+				'\n\n' +
+				'To avoid an image build error, the CLI has disabled architecture selection. ' +
+				'However, this means that the architecture of the machine where the Docker ' +
+				'Engine is running will be used by default to select the architecture ' +
+				'which may result in "exec format error" at runtime. To avoid this problem, ' +
```
klutchell, line 355:
```suggestion
				'which may result in "exec format error" at runtime. To avoid this warning, ' +
```

lib/build.ts
```diff
@@ -187,3 +194,169 @@ export async function runBuildTask(
...
+				'error" at runtime. To avoid this problem, please update or replace the images ' +
+				'that do not support architecture selection.',
+		);
+	} else {
+		// Only old v1 images
+		console.warn(
+			'Warning: Dockerfile contains a images do not support architecture selection.\n-' +
+				v1Images.join('\n-') +
+				'\n\n' +
+				'To avoid an image build error, the CLI has disabled architecture selection. ' +
```
klutchell, line 352:
```suggestion
				'To avoid an image build error, the CLI has disabled platform selection. ' +
```

klutchell: Suggested rewording:
```
'To avoid a possible build error, the CLI has disabled platform selection. ' +
'As a result, the architecture of the machine where the Docker Engine ' +
'is running will be used to select the architecture when pulling upstream images' +
'which may result in "exec format error" at runtime. To avoid this warning, ' +
'update or replace the images that do not support platform selection.',
```

lib/build.ts
klutchell: Suggested rewording:
```
'To avoid a possible build error, the CLI has disabled platform selection. ' +
'As a result, the architecture of the machine where the Docker Engine ' +
'is running will be used to select the architecture when pulling upstream images' +
'which may result in "exec format error" at runtime. To avoid this warning, ' +
'update or replace the images that do not support platform selection.',
```

klutchell: I actually dropped a `please` from the text because it seemed to wordy. How un-Canadian of me.

lib/build.ts
toochevere: Well shoot... this presents two problems

1. There doesn't appear to be a good library for parsing the FROM line (at least, as far as my looking around is concerned, I haven't found one).  But I can still do a (hacky?) look through the parameters and make sure I am actually getting the image.  This doesn't seem like a terrible option.   
2. It did not occur to me previously that a user might specify `platform` here.  How should we handle this?   Going to [FD to discuss](https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc)

toochevere: Talking with @klutchell on FD, `platform` on the command line is ignored when it is specified in the dockerfile.  So we will just let the engine decide what to do, ignoring it in the dockerfile.  (regarding 2nd point)

lib/build.ts
toochevere: The error was something suggested by an earlier reviewer.  But I think I agree that it's a bit draconian.

toochevere: Given that we are doing this, we probably don't need to look for `balenalib` images at all.  V1 images will all be treated to the same warning.

lib/build.ts
```diff
@@ -187,3 +194,173 @@ export async function runBuildTask(
...
+//   So what do we do?
+//
+//      + In the case of all images having a V2 manifest, we can simply pass the platform flag.
+//
+//      + In the case of having some V1 manifests ...
+//          - If we have balenalib images, we know that they have been updated.  Throw an error
+//            and tell the user to update
+//          - If we do not have balenalib images, don't pass the platform flag.  Let Docker
+//            close its eyes and hope for the best.  This opens the possibility for an exec
+//            error, but also allows users to continue to use v1 images if they want to / need to.
```
klutchell, line 248: We've removed the special cases around balenalib images now right? I think this comment needs updating.

lib/build.ts
```diff
@@ -187,3 +194,173 @@ export async function runBuildTask(
...
+			if (fromLineArguments.length === 0) {
+				// bail out, no image specified?
+				return;
+			}
+
+			let imgRef = fromLineArguments[0].getValue();
+
+			// hacky way to ignore the --platform flag in the FROM line
+			// In that case, just get the next argument :-p
+			if (imgRef.startsWith('--platform')) {
```
klutchell, line 306: At first I thought this would break if the `--platform=linux/arm/v7` was replaced with `--platform linux/arm/v7` but the equals seems to be required according to the docker daemon.
```
FROM --platform linux/arm/v7 alpine:3.12.8
```
```
Error response from daemon: dockerfile parse error line 1: FROM requires either one or three arguments
```
What you have here should work albeit a little hacky as you said.

lib/build.ts
```diff
@@ -187,3 +194,173 @@ export async function runBuildTask(
...
+		}),
+	);
+
+	if (v1Images.length === 0) {
+		// Only v2 images, let Docker receive `--platform`
+		return true;
+	} else if (v2Images.length > 0) {
+		// Mixed v1 and v2 images
+		console.log(
+			'Warning: Dockerfile contains a mixes images that require an architecture selection with ones that do not support it.\n' +
```
klutchell, line 337:
```suggestion
			'Warning: Dockerfile contains a mixes images that require a platform selection with ones that do not support it.\n' +
```

lib/build.ts
```diff
@@ -187,3 +194,173 @@ export async function runBuildTask(
...
+				'The following images do not support architecture selection:\n-' +
+				v1Images.join('\n-') +
+				'\n\n' +
+				'The following images require architecture selection:\n-' +
```
klutchell, line 341:
```suggestion
				'The following images require platform selection:\n-' +
```

lib/build.ts
```diff
@@ -187,3 +194,173 @@ export async function runBuildTask(
...
+				v2Images.join('\n-') +
+				'\n\n' +
+				'To avoid a possible build error, the CLI has disabled platform selection. ' +
+				'As a result, the architecture of the machine where the Docker Engine ' +
```
klutchell, line 345: IMO this is the only place where it is appropriate to use `architecture` as it is a classification of the workstation, everywhere else we should use `platform` to align with Docker vernacular.

toochevere: Ok, I think I have all the cases sorted out now.

klutchell:
Looks good to me, but @robertgzr will have to sign off on his requested changes either way.

toochevere:
@balena-ci I self certify!

toochevere:
Closing this PR because @robertgzr is on vacay until the 27th.  Will open a fresh one and link back so I can merge.

PR #93, opened 2021-08-18
Pass additional build options to docker

robertgzr:
This pulls in resin-compose-parse version TODO which supports
build options cache_from, extra_hosts, isolation, network, shm_size and
target.

Change-type: minor


lib/utils.ts
```diff
@@ -59,10 +59,36 @@ export function generateBuildTasks(
...
+					'target',
+				),
+				// pass through build options from composition
+				// translating to dockerode ImageBuildOptions properties
+				{
+					dockerOpts: _.merge(
+						{},
+						img.image.cache_from ? { cachefrom: img.image.cache_from } : {},
+						img.image.network ? { networkmode: img.image.network } : {},
+						img.image.shm_size ? { shmsize: img.image.shm_size } : {},
```
robertgzr, line 85: let's continue the conversation from https://github.com/balena-io-modules/resin-compose-parse/pull/59#pullrequestreview-733132850

lib/utils.ts
```diff
@@ -59,10 +59,36 @@ export function generateBuildTasks(
...
+					'shm_size',
+					'target',
+				),
+				// pass through build options from composition
+				// translating to dockerode ImageBuildOptions properties
+				{
+					dockerOpts: _.merge(
+						{},
+						img.image.cache_from ? { cachefrom: img.image.cache_from } : {},
+						img.image.network ? { networkmode: img.image.network } : {},
```
robertgzr, line 84: let's continue the conversation from https://github.com/balena-io-modules/resin-compose-parse/pull/59#pullrequestreview-733128487

robertgzr:
marking this as ready for review, since I consider https://github.com/balena-io-modules/resin-compose-parse/pull/59 done at this point

robertgzr:
rebased on master and changed the temp commit to refer to latest HEAD on the compose-parse branch

lib/utils.ts
```diff
@@ -59,10 +59,35 @@ export function generateBuildTasks(
...
+				// We also remove anything that goes into dockerOpts in the next step
+				_.omit(
+					img.image,
+					'image',
+					'cache_from',
+					'extra_hosts',
+					'network',
+					'shm_size',
+					'target',
+				),
```
karaxuna: No need to use lodash, you could do:

```typescript
const {
  image,
  cache_from: cachefrom,
  extra_hosts: extrahosts,
  network: networkmode,
  shm_size: shmsize,
  target,
  ...restImageProps
} = img.image
```

`restImageProps` is an object containing all props except keys above. You could use `target`, `shmsize`... in code below.

lib/utils.ts
```diff
@@ -59,10 +59,35 @@ export function generateBuildTasks(
...
+				// pass through build options from composition
+				// translating to dockerode ImageBuildOptions properties
+				{
+					dockerOpts: _.merge(
+						{},
+						img.image.cache_from ? { cachefrom: img.image.cache_from } : {},
```
karaxuna: Continuing comment above, you would:

```typescript
cachefrom ? { cachefrom } : {}
```

lib/utils.ts
```diff
@@ -47,6 +47,19 @@ export function generateBuildTasks(
 			if (img.image.context == null) {
 				throw new Error('Must have a context specified with a Dockerfile');
 			}
+			// It's possible to specify an image name as well as
+			// a build, but that doesn't make sense in a balena
+			// ecosystem, so we remove it.
+			// We also remove anything that goes into dockerOpts
+			const {
+				image,
```
pdcastro: Build is failing on this line:
```
lib/utils.ts:55:5 - error TS2339: Property 'image' does not exist on type 'BuildConfig'.

55     image,
       ~~~~~
```
Because, I think, `img.image.image` is not a string here, it's a `BuildConfig` object that indeed does not have a property named `image`:
```
export interface BuildConfig {
    args?: Dict<string>;
    cache_from?: ListOfUniqueItems<string>;
    context: string;
    dockerfile?: string;
    extra_hosts?: ListOfUniqueItems<string>;
    isolation?: string;
    labels?: Dict<string>;
    network?: string;
    shm_size?: number | string;
    tag?: string;
    target?: string;
}
```

lib/utils.ts
```diff
@@ -59,10 +72,21 @@ export function generateBuildTasks(
...
-				// It's possible to specify an image name as well as
-				// a build, but that doesn't make sense in a balena
-				// ecosystem, so we remove it
-				_.omit(img.image, 'image'),
+				// Pass through args, context, labels, tag.
+				imageProps,
+				// Pass through build options from composition
+				// translating to dockerode ImageBuildOptions properties
+				{
+					dockerOpts: _.merge(
```
pdcastro: To confirm, the purpose of these changes to the `generateBuildTasks` function is to add these properties to `dockerOpts`, right? Is this function the best place to do this though? Maybe it is, but I would have thought that the `runBuildTask` function, where `dockerOpts` is already being modified, would be a better place, because then most handling of `dockerOpts` is done in the same place:
* https://github.com/balena-io-modules/balena-multibuild/blob/v4.12.2/lib/build.ts#L138-L144
* https://github.com/balena-io-modules/balena-multibuild/blob/v4.12.2/lib/build.ts#L164-L188

I think that changing `dockerOpts` in `runBuildTask` has fewer implications on existing logic in `balena-cli` and `balena-builder` (logic that makes assumptions about the "contents" of `dockerOpts` after a call to `fromImageDescriptors`), because `runBuildTask` makes those changes right before sending the Docker API request to the Docker engine.

robertgzr: tbh I'm not sure... I don't see how to get the `ImageDescriptor` from `runBuildTask`, since that operates solely on `BuildTask`, so I need to put those options there. I guess I could create new fields, maybe `target` and `buildOpts`. they seem redundant with `dockerOpts` also there though...

robertgzr: also, we would then have to reconcile the situation where somebody might pass `network_mode: host` via `BuilTask.dockerOpts` while another value is present in the Composition

pdcastro: Fair enough, somehow I thought that `ImageDescriptor` was referenced from a `BuildTask`, but it is not. Sorry for the noise!

robertgzr: @pdcastro do you know if we need to worry about this, at least for the cli


> think that changing dockerOpts in runBuildTask has fewer implications on existing logic in balena-cli and balena-builder (logic that makes assumptions about the "contents" of dockerOpts after a call to fromImageDescriptors),

cc @toochevere for the builder

pdcastro: Yes, meaning simply that we should check the source code of `balena-cli` and `balena-builder` for how they use `resin-multibuild` and in particular how they set `dockerOpts`. This should be done alongside testing this PR's integration with `balena-cli` and `balena-builder`, before this PR is merged, by creating draft PRs against `balena-cli` and `balena-builder` and using the draft `resin-multibuild` npm package version that balena CI created for this PR, which can be seen in the 'versions' tab at https://www.npmjs.com/package/resin-multibuild

E.g. in the `package.json` of `balena-cli` and `balena-builder`, we would replace:
```json
"resin-multibuild": "^4.12.2",
```
with:
```json
"resin-multibuild": "4.13.0-rgz-pass-build-options-18af5273d0396ada971a8ca8b2b9303781b71ef8",
```
And then check that the automated tests pass for `balena-cli` and `balena-builder`, and do some manual testing of the new features that are made available by this PR.
After we confirm that all CLI and builder tests pass, and the new features work as expected, this PR can be merged and then the `balena-cli` and `balena-builder` PRs can be updated with the final, non-draft version of `resin-multibuild`.

dfunckt:
@robertgzr please replace `resin-compose-parse` with `@balena/compose-parse@v3.0.0`

lib/utils.ts
robertgzr: ~~allow for local builds~~ allow

lib/utils.ts
robertgzr: ~~allow for local builds~~ disallow

lib/utils.ts
```diff
@@ -79,7 +84,6 @@ export function generateBuildTasks(
 					dockerOpts: _.merge(
 						{},
 						cachefrom ? { cachefrom } : {},
-						networkmode ? { networkmode } : {},
```
pdcastro: ~~This commit's title is "Disallow networkmode build option" and [...]~~

pdcastro: Nope, false alarm. I was mixing buildargs with buildopts.

lib/utils.ts
```diff
@@ -59,10 +76,20 @@ export function generateBuildTasks(
...
+				// Pass through build options from composition
+				// translating to dockerode ImageBuildOptions properties
+				{
+					dockerOpts: _.merge(
+						{},
+						cachefrom ? { cachefrom } : {},
+						shmsize ? { shmsize } : {},
+						target ? { target } : {},
+						extrahosts ? { extrahosts } : {},
+					),
```
pdcastro: > `extrahosts` _string_ Extra hosts to add to /etc/hosts
> https://docs.docker.com/engine/api/v1.41/#operation/ImageBuild

What a peculiar image build option! 😀 

As these options don't overlap with each other, I believe JS's builtin spread syntax would be sufficient as follows, but it does not really matter:
```suggestion
					dockerOpts: {
						...(cachefrom ? { cachefrom } : {}),
						...(shmsize ? { shmsize } : {}),
						...(target ? { target } : {}),
						...(extrahosts ? { extrahosts } : {}),
					},
```

lib/utils.ts
```diff
@@ -59,10 +76,20 @@ export function generateBuildTasks(
...
+				},
+				// TODO dockerPlatform (img.platform) ?
```
pdcastro: `img.platform` would presumably override architecture selection from the balena fleet's default architecture?
Yes, "to do" for some other PR...

lib/utils.ts
```diff
@@ -47,6 +47,23 @@ export function generateBuildTasks(
 			if (img.image.context == null) {
 				throw new Error('Must have a context specified with a Dockerfile');
 			}
+			// It's possible to specify an image name as well as
+			// a build, but that doesn't make sense in a balena
+			// ecosystem, so we remove it.
```
pdcastro: These 3 comment lines refer to a deleted line of code:
```js
_.omit(img.image, 'image'),
```
... which used to select all properties of `img.image` except for `img.image.image`. This is confusing but if I am getting it right (I may well be wrong!), it refers to this documented feature:

> https://docs.docker.com/compose/compose-file/compose-file-v2/#build
> [...] If you specify `image` as well as `build`, then Compose names the built image with the `webapp` and optional `tag` specified in `image`:
> ```yaml
> build: ./dir
> image: webapp:tag
> ```
> This results in an image named `webapp` and tagged `tag`, built from `./dir`.

Such an `img.image.image` string property, which was previously omitted, after this PR becomes part of `...imageProps` (no longer omitted), which may have been unintentional. The "fix" in this sense would be easy, probably just adding an `image` line above `imageProps`:

```js
			const {
				cache_from: cachefrom,
				extra_hosts: extrahosts,
				shm_size: shmsize,
				target,
				image,
				...imageProps
			} = img.image
```
(Although the linter might complain about an unused `image` variable.)

BUT, I would also challenge / consider the comment's statement that _"that doesn't make sense in a balena ecosystem."_ There is issue balena-io/balena-cli/issues/1874 where users (including our very @joshbwlng) actually wanted `balena build` or `balena push` to a local device to name the image built for a service using the `img.image.image` 'name:tag'. In the case of `balena build`, something related is currently achieved with the `--projectName` and `--tag` flags:

```
$ balena help build
  -n, --projectName <projectName>
      Name prefix for locally built images. This is the 'projectName' portion
      in 'projectName_serviceName:tag'. The default is the directory name.

  -t, --tag <tag>
      Tag locally built Docker images. This is the 'tag' portion
      in 'projectName_serviceName:tag'. The default is 'latest'.
```

These flags have the limitation that they apply globally to all services in `docker-compose.yml`, however `img.image.image` would have the ability to set image names for individual services.

Anyway, fixing balena-io/balena-cli/issues/1874 is probably outside the scope of this PR, but I think this PR needs to at least amend those 3 comment lines (perhaps replace "doesn't make sense" with "todo: allow image property and resolve balena-io/balena-cli/issues/1874") and ensure that `img.image.image` property continues to be omitted (as before this PR) if we don't want to fix that issue or implement a new feature here.

robertgzr: > Anyway, fixing balena-io/balena-cli/issues/1874 is probably outside the scope of this PR, but I think this PR needs to at least amend those 3 comment lines (perhaps replace "doesn't make sense" with "todo: allow image property and resolve balena-io/balena-cli/issues/1874") and ensure that img.image.image property continues to be omitted (as before this PR) if we don't want to fix that issue or implement a new feature here.

it looks like this would require a compose-parse change... currently the ImageDescriptor.image field is defined as [`string | BuildConfig`](https://github.com/balena-io-modules/balena-compose-parse/blob/master/src/types.ts#L220)

so we can't pass along the image property *and* the build config, I wonder what the `omit` was trying to accomplish? since `BuildConfig` doesn't have an `image` field...

robertgzr: I added the link to this issue anyway... might be worth opening a compose-parse ticket with this as well? although I feel like this falls into the category of behaviour that would differ between the builder and the cli (which we don't want - according to recent discussions)

lib/utils.ts
```diff
@@ -47,6 +47,24 @@ export function generateBuildTasks(
...
+			// TODO: should we allow image property to resolve:
+			// https://github.com/balena-io/balena-cli/issues/1874
```
pdcastro: To me, this TODO comment would belong here if the code below was explicitly removing the `image` field like the deleted `_.omit(img.image, 'image')` line used to do, or like I had suggested the code to be amended. As the code was not amended, I think we are better off deleting this TODO comment. The reminder of the issue is the issue itself (balena-io/balena-cli/issues/1874).

lib/utils.ts
pdcastro: > so we can't pass along the image property _and_ the build config, I wonder what the `omit` was trying to accomplish? since `BuildConfig` doesn't have an `image` field...

`omit` was trying to accomplish what the comment states:

```ts
	// It's possible to specify an image name as well as
	// a build, but that doesn't make sense in a balena
	// ecosystem, so we remove it.
```

It could have historical reasons, including the fact that the much of the CLI and builder codebase were written in Coffeescript or untyped Javascript (CLI and builder), and `multibuild` could not make assumptions based on Typescript interfaces. Event today, the CLI is not 100% Typescript (almost there). The statement that _"we can't pass along the image property `and` the build config"_ assumes that this is somehow enforced or validated, presumably through Typescript interfaces and/or the logic of `balena-io-modules/balena-compose-parse`. Nowadays, I think this is fair enough.

The analysis seems to be that, even before this PR (current master branch), `_.omit(img.image, 'image'),` had no reason to be, and therefore the line can be safely deleted by this PR. I cannot disprove it, I can only highlight that the logic has changed. :-)

I think you will agree that, having deleted `_.omit(img.image, 'image')` (and not replaced it with other code to the same effect), it does not make sense to keep those 3 lines of comment that describe the line that was deleted, and it is factually incorrect for the comment to state that the `image` field is removed ("`so we remove it`") when the code that did the removal was removed.

In this case, I'd say, delete the 3 lines of comment as well.

PR #94, opened 2021-09-16
Determine when to use the docker 'platform' flag to enable multiarch support

toochevere:
This change is part of the larger effort to enable mult-arch image support.

Change-type: minor
Signed-off-by: Paul Jonathan <pj@balena-io>
See: https://github.com/balena-io-modules/balena-multibuild/pull/92


toochevere:
@balena-ci I self certify!

PR #96, opened 2021-09-17
Add Logger to BuildTask interface, use logger for multiarch warnings

pdcastro:
Change-type: patch

### DIFF
--- a/lib/build.ts
+++ b/lib/build.ts
@@ -250,6 +250,9 @@ async function checkAllowDockerPlatformHandling(
 ): Promise<boolean> {
 	let imageReferences: string[];
 
+	const debug = task.logger ? task.logger.debug : () => undefined;
+	const warn = task.logger ? task.logger.warn : () => undefined;
+
 	const checkHasPlatformInfo = async (repository: string) => {
 		try {
 			const manifest = await getManifest(docker.modem, repository);
@@ -258,8 +261,8 @@ async function checkAllowDockerPlatformHandling(
 			// eat the exception... yummy!
 		}
 
-		console.log(
-			`Image manifest data unavailable for ${repository}. No support for specifying platform.`,
+		debug(
+			`${task.serviceName}: Image manifest data unavailable for ${repository}`,
 		);
 		// at this time, treat any errors as false (no platform support)
 		return false;
@@ -270,7 +273,9 @@ async function checkAllowDockerPlatformHandling(
 	} else {
 		if (!task.dockerfile) {
 			// Sanity check
-			console.log('Build task does not have an associated Dockerfile');
+			debug(
+				`${task.serviceName}: Build task does not have an associated Dockerfile`,
+			);
 			return false;
 		}
 		const { DockerfileParser } = await import('dockerfile-ast');
@@ -281,9 +286,7 @@ async function checkAllowDockerPlatformHandling(
 		);
 		if (fromInstructions.length === 0) {
 			// Sanity check
-			console.log(
-				'Build task associated Dockerfile does no reference any images',
-			);
+			debug(`${task.serviceName}: Dockerfile does not reference any images`);
 			return false;
 		}
 		imageReferences = [];
@@ -327,33 +330,26 @@ async function checkAllowDockerPlatformHandling(
 		// All images specify platform, let Docker receive `--platform`
 		return true;
 	} else if (imagesWithPlatformSupport.length > 0) {
-		// Mixed images
-		console.log(
-			'Warning: Dockerfile contains a mixes images that require a platform selection with ones that do not support it.\n' +
-				'The following images do not support platform selection:\n-' +
-				imagesWithoutPlatformSupport.join('\n-') +
-				'\n\n' +
-				'The following images require platform selection:\n-' +
-				imagesWithPlatformSupport.join('\n-') +
-				'\n\n' +
-				'To avoid a possible build error, the CLI has disabled platform selection. ' +
-				'As a result, the architecture of the machine where the Docker Engine ' +
-				'is running will be used to select the platform when pulling upstream images ' +
-				'which may result in "exec format error" at runtime. To avoid this warning, ' +
-				'update or replace the images that do not support platform selection.',
-		);
-	} else {
-		// Only images without platform support
-		console.log(
-			'Warning: Dockerfile contains a images do not support platform selection.\n-' +
-				imagesWithoutPlatformSupport.join('\n-') +
-				'\n\n' +
-				'To avoid a possible build error, the CLI has disabled platform selection. ' +
-				'As a result, the architecture of the machine where the Docker Engine ' +
-				'is running will be used to select the platform when pulling upstream images ' +
-				'which may result in "exec format error" at runtime. To avoid this warning, ' +
-				'update or replace the images that do not support platform selection.',
-		);
+		// Here we know that the service references at least 2 images
+		// (at least 1 with platform support and at least 1 without platform
+		// support). Therefore the service is not an "external image service"
+		// (`task.imageName` must be falsy), and it must have a multi-stage
+		// Dockerfile that references at least 2 images.
+		warn(`\
+Service '${task.serviceName}':
+  Multi-stage Dockerfile found with a mix of base images that require
+  CPU architecture selection and base images that do not support it.
+  The following base images do not support CPU architecture selection:
+  - ${imagesWithoutPlatformSupport.join('\n  - ')}
+  The following base images require CPU architecture selection:
+  - ${imagesWithPlatformSupport.join('\n  - ')}
+  As a result, the CPU architecture of the machine where the Docker Engine
+  is running will be used by default to select base images that require
+  architecture selection. This may result in incorrect architecture selection
+  and "exec format error" at runtime. It is usually possible to override the
+  architecture in the FROM line with e.g. "FROM --platform=linux/arm/v7",
+  or by adding the sha256 digest of the image for a specific architecture
+  with e.g. "FROM debian@sha256:094f57...".`);
 	}
 
 	// Proceeding after warnings.  Docker will _not_ receive `--platform`
--- a/lib/build.ts
+++ b/lib/build.ts
@@ -248,7 +248,7 @@ async function checkAllowDockerPlatformHandling(
 	task: BuildTask,
 	docker: Dockerode,
 ): Promise<boolean> {
-	let imageReferences: string[];
+	const imageReferences: string[] = [];
 
 	const debug = task.logger ? task.logger.debug : () => undefined;
 	const warn = task.logger ? task.logger.warn : () => undefined;
@@ -269,7 +269,7 @@ async function checkAllowDockerPlatformHandling(
 	};
 
 	if (task.imageName) {
-		imageReferences = [task.imageName + !!task.tag ? `:${task.tag}` : ''];
+		imageReferences.push(task.imageName);
 	} else {
 		if (!task.dockerfile) {
 			// Sanity check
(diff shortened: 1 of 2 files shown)


lib/build.ts
```diff
@@ -328,32 +333,21 @@ async function checkAllowDockerPlatformHandling(
...
-		// Only images without platform support
-		console.log(
-			'Warning: Dockerfile contains a images do not support platform selection.\n-' +
-				imagesWithoutPlatformSupport.join('\n-') +
-				'\n\n' +
-				'To avoid a possible build error, the CLI has disabled platform selection. ' +
-				'As a result, the architecture of the machine where the Docker Engine ' +
-				'is running will be used to select the platform when pulling upstream images ' +
-				'which may result in "exec format error" at runtime. To avoid this warning, ' +
-				'update or replace the images that do not support platform selection.',
```
pdcastro, line 355: See https://github.com/balena-io/balena-cli/pull/2301#pullrequestreview-757537617 for the reasoning for removing this `else` branch.

lib/build.ts
```diff
@@ -328,32 +333,21 @@ async function checkAllowDockerPlatformHandling(
...
+-${imagesWithoutPlatformSupport.join('\n-')}
+
+The following base images require CPU architecture selection:
+-${imagesWithPlatformSupport.join('\n-')}
+
+As a result, the CPU architecture of the machine where the Docker Engine
+is running will be used to select base images that require architecture
+selection. This may result in incorrect architecture selection and "exec
+format error" at runtime. To avoid this warning, update or replace base
+images that do not support architecture selection.`);
```
pdcastro, line 350: I haven't tested this yet, not sure it looks good. I will test it with balena-io/balena-cli/pull/2301 .

lib/build.ts
```diff
@@ -328,32 +333,19 @@ async function checkAllowDockerPlatformHandling(
...
+  CPU architecture selection and base images that do not support it.
+  The following base images do not support CPU architecture selection:
+  - ${imagesWithoutPlatformSupport.join('\n  - ')}
+  The following base images require CPU architecture selection:
+  - ${imagesWithPlatformSupport.join('\n  - ')}
+  As a result, the CPU architecture of the machine where the Docker Engine
+  is running will be used to select base images that require architecture
+  selection. This may result in incorrect architecture selection and "exec
+  format error" at runtime. To avoid this warning, update or replace base
+  images that do not support architecture selection.`);
```
pdcastro, line 348: Here's how this ends up looking with the CLI -- the lines prefixed with `[Warn]`:
```
$ ./bin/balena build t -A armv7hf -d raspberrypi3 --nocache --pull --debug
...
[Info]    Building for armv7hf/raspberrypi3
[Info]    Docker Desktop detected (daemon architecture: "x86_64")
[Info]      Docker itself will determine and enable architecture emulation if required,
[Info]      without balena-cli intervention and regardless of the --emulated option.
[Debug]   Found build tasks:
[Debug]       main: build [.]
[Debug]   Resolving services with [raspberrypi3|armv7hf]
[Debug]   Found project types:
[Debug]       main: Standard Dockerfile
[Debug]   Prepared tasks; building...
[Warn]    Service 'main':
[Warn]      Multi-stage Dockerfile found with a mix of base images that require
[Warn]      CPU architecture selection and base images that do not support it.
[Warn]      The following base images do not support CPU architecture selection:
[Warn]      - balenalib/raspberrypi3-node:14-buster-run
[Warn]      The following base images require CPU architecture selection:
[Warn]      - debian:10
[Warn]      As a result, the CPU architecture of the machine where the Docker Engine
[Warn]      is running will be used to select base images that require architecture
[Warn]      selection. This may result in incorrect architecture selection and "exec
[Warn]      format error" at runtime. To avoid this warning, update or replace base
[Warn]      images that do not support architecture selection.
[Build]   main Step 1/5 : FROM balenalib/raspberrypi3-node:14-buster-run as base
[Build]   main  ---> e074e38a89ba
```

lib/build.ts
```diff
@@ -328,32 +333,19 @@ async function checkAllowDockerPlatformHandling(
...
+  CPU architecture selection and base images that do not support it.
+  The following base images do not support CPU architecture selection:
+  - ${imagesWithoutPlatformSupport.join('\n  - ')}
+  The following base images require CPU architecture selection:
+  - ${imagesWithPlatformSupport.join('\n  - ')}
+  As a result, the CPU architecture of the machine where the Docker Engine
+  is running will be used to select base images that require architecture
+  selection.  This may result in incorrect architecture selection and
+  "exec format error" at runtime. To avoid this warning, update or replace
+  base images that do not support architecture selection.`);
```
pdcastro, line 348: Here's how this ends up looking with the CLI -- the lines prefixed with `[Warn]`:
```
$ ./bin/balena build t -A armv7hf -d raspberrypi3 --nocache --pull --debug
...
[Info]    Building for armv7hf/raspberrypi3
[Info]    Docker Desktop detected (daemon architecture: "x86_64")
[Info]      Docker itself will determine and enable architecture emulation if required,
[Info]      without balena-cli intervention and regardless of the --emulated option.
[Debug]   Found build tasks:
[Debug]       main: build [.]
[Debug]   Resolving services with [raspberrypi3|armv7hf]
[Debug]   Found project types:
[Debug]       main: Standard Dockerfile
[Debug]   Prepared tasks; building...
[Warn]    Service 'foo':
[Warn]      Multi-stage Dockerfile found with a mix of base images that require
[Warn]      CPU architecture selection and base images that do not support it.
[Warn]      The following base images do not support CPU architecture selection:
[Warn]      - balenalib/raspberrypi3-node:14-buster-run
[Warn]      The following base images require CPU architecture selection:
[Warn]      - debian:10
[Warn]      - busybox
[Warn]      As a result, the CPU architecture of the machine where the Docker Engine
[Warn]      is running will be used to select base images that require architecture
[Warn]      selection.  This may result in incorrect architecture selection and
[Warn]      "exec format error" at runtime. To avoid this warning, update or replace
[Warn]      base images that do not support architecture selection.
[Build]   main Step 1/5 : FROM balenalib/raspberrypi3-node:14-buster-run as base
[Build]   main  ---> e074e38a89ba
```

pdcastro: Regarding the advice:
```
To avoid this warning, update or replace base images that do not support architecture selection.
```
Ideally the latest balenalib base images would qualify as base images that support architecture selection, otherwise we are effectively telling balena users to replace balenalib base images with non-balenalib base images.

I have gathered that `balenalib/raspberrypi3-node:latest` is treated by `balena-multibuild` as not supporting platform selection.

I see that [a discussion started in flowdock](https://www.flowdock.com/app/rulemotion/i-cli/threads/RuSu1KiWOn62xaGy7O2sn8m8BUc) with @nghiant2710, who appears to claim that the latest balenalib base images correctly declare their platform or were generated as they should. Well, either `balena-multibuild` is not correctly detecting this, or balenalib base images are still missing something...

pdcastro: I have just tested `FROM --platform` with this multi-stage `Dockerfile.template`:
```Dockerfile
FROM balenalib/%%BALENA_MACHINE_NAME%%-node as base
FROM --platform=linux/arm/v7 busybox as base2
FROM --platform=linux/arm/v7 debian:10
COPY --from=base /usr/local/bin/node /tmp/node
COPY --from=base2 /bin/busybox /tmp/busybox
CMD ["/bin/bash"]
```

And `balena build -A armv7hf -d raspberrypi3`, thus meaning to build for ARM, but running on an Intel laptop. The CLI prints the mixed-images warning as expected. The expected, but still interesting observation is that:

* Without `FROM --platform=linux/arm/v7` (no FROM-line platform selection), the `busybox` and `debian` images are selected for Intel, as the warning warns. The templated `rasbperrypi3-node` base image is for ARM, of course, as it is a single-arch image.
* With `FROM --platform=linux/arm/v7`, the `busybox` and `debian` images are selected for ARM v7, also as expected.

Therefore, instead of advising `To avoid this warning, update or replace base images that do not support architecture selection` (especially if this was not yet feasible with `balenalib`), we could instead advise the possibility of using `FROM --platform`, if users needed to mix base images. Perhaps:

```ts
  `As a result, the CPU architecture of the machine where the Docker Engine
  is running will be used by default to select base images that require
  architecture selection. This may result in incorrect architecture selection
  and "exec format error" at runtime. It is usually possible to override the
  architecture in the FROM line with e.g. "FROM --platform=linux/arm/v7",
  or by adding the sha256 digest of the image for a specific architecture
  with e.g. "FROM debian@sha256:094f5..."`);
```

Also:
* As we know the target architecture (`-A armv7hf -d raspberrypi3`),
* And we know we cannot add a common `platform` option to the Docker API (as Docker Engine will throw an error, and the reason why we print the mixed-image warning),
* And we know which images require architecture selection (we are printing the list in the warning message),
* And we know that `FROM --platform=linux/arm/v7` would solve the problem (I'm planning to add this advice to the warning message),
* Then (in another PR) `balena-multibuild` could automatically and on-the-fly add `FROM --platform=linux/arm/v7` where needed (in the build stream only, not saved to disk).

lib/build.ts
```diff
@@ -258,19 +261,21 @@ async function checkAllowDockerPlatformHandling(
...
+		debug(
+			`${task.serviceName}: Image manifest data unavailable for ${repository}`,
 		);
 		// at this time, treat any errors as false (no platform support)
 		return false;
 	};
 
 	if (task.imageName) {
-		imageReferences = [task.imageName + !!task.tag ? `:${task.tag}` : ''];
+		imageReferences.push(task.imageName);
```
pdcastro, line 272: This is a bug fix:
* `task.imageName` is the name of an "external image" when the `docker-compose.yml` file contains, for example, an `image: debian:10` line for a service, instead of a `build:` line/section. The value of `task.imageName` is something like `debian:10`, including the tag part.
* `task.tag` is used to tag the local images that are built (with a Dockerfile) on the machine where Docker is running. For a project name 'myFleet' and a service name 'myService', the value of `task.tag` would be `myfleet_myservice`.

So, in this example, putting together `task.imageName` and `task.tag` would result in `debian:10:myfleet_myservice`, which isn't right. Except that there was also a bug within the bug, in that:
```
task.imageName + !!task.tag ? `:${task.tag}` : ''
```
associates as:
```
(task.imageName + !!task.tag) ? `:${task.tag}` : ''
```
rather than the semantically intended:
```
task.imageName + (!!task.tag ? `:${task.tag}` : '')
```

lib/build.ts
```diff
@@ -281,32 +286,18 @@ async function checkAllowDockerPlatformHandling(
...
-			if (imgRef.startsWith('--platform')) {
-				if (fromLineArguments.length < 2) {
-					// bail out, no image specified?
-					return;
+		for (const inst of fromInstructions) {
+			for (const arg of inst.getArguments()) {
+				const val = arg.getValue();
+				if (!val.startsWith('--')) {
+					imageReferences.push(val);
+					break;
```
pdcastro, line 297: There was no bug here, but the benefits of this change are:
* Future proof against other FROM `--` options that may be added in the future, besides `--platform`.
* Fewer lines of code and, in my biased opinion, easier to read.

lib/build.ts
```diff
@@ -327,33 +318,26 @@ async function checkAllowDockerPlatformHandling(
...
+  - ${imagesWithoutPlatformSupport.join('\n  - ')}
+  The following base images require CPU architecture selection:
+  - ${imagesWithPlatformSupport.join('\n  - ')}
+  As a result, the CPU architecture of the machine where the Docker Engine
+  is running will be used by default to select base images that require
+  architecture selection. This may result in incorrect architecture selection
+  and "exec format error" at runtime. It is usually possible to override the
+  architecture in the FROM line with e.g. "FROM --platform=linux/arm/v7",
+  or by adding the sha256 digest of the image for a specific architecture
+  with e.g. "FROM debian@sha256:094f57...".`);
```
pdcastro, line 340: For another PR, we could detect whether the user is already following this advice to avoid printing the warning, i.e. platform-ready images in the Dockerfile are already either selecting the image with a sha256 digest (as support agents have been advising users to do for years, as a workaround for lack of multiarch support), or already using `FROM --platform`. It can be confusing to be warned to do something (use `FROM --platform`) when you are already doing it.

lib/build.ts
toochevere: thank you so much for catching this!

lib/build.ts
toochevere: Post merge task:

- [x] Add an issue to address this

toochevere (approved):

lib/build.ts
toochevere: See: https://github.com/balena-io-modules/balena-multibuild/issues/97

lib/build.ts
toochevere: see:  https://github.com/balena-io-modules/balena-multibuild/issues/97

PR #99, opened 2021-11-10
Inject template variables as build args

### LINKED ISSUE #67: Use ARG environment variable as template system
Hi! 

I'm in the Barcelona 2019 Hackaton, I've been speaking with some of your developers and I've been advised to post this issue here. 

Regarding the template system implemented in your platform, to be able to build Docker images for several boards, i. e. Raspberry Pi 2 or 3 at the same time, documented here: https://www.balena.io/docs/learn/develop/dockerfile/; I believe it should be fully compatible with the new standard ARG environments variables in FROM statements during the build time of a docker image. Which is documented here: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

Take into account the example Dockerfile given by this repo: https://github.com/balena-io-playground/balena-sense-hat-python-starter/blob/master/Dockerfile

```
FROM resin/raspberrypi3-debian:latest
```

This FROM statement works locally and remotely for Raspberry Pi 3 images, if we want to use this Dockerfile  in different machines, we should implement your template system: 

```
FROM resin/%%BALENA_MACHINE_NAME%%-debian:latest
``` 
And consequently rename `Dockerfile` to `Dockerfile.template`, that would make this Dockerfile not complience with Docker standards. 

One would expect that `BALENA_MACHINE_NAME` would be present as a environment variable as build time during your building, thus, 

```
ARG BALENA_MACHINE_NAME
FROM resin/${BALENA_MACHINE_NAME}-debian:latest
``` 

would work, but it does not! The best approach I've currently found it is something like this:

```
ARG BALENA_MACHINE_NAME=%%BALENA_MACHINE_NAME%%
FROM resin/${BALENA_MACHINE_NAME}-debian:latest
``` 
Using a `Dockerfile.template` as a Dockerfile and it is both compatible with Docker cli and your template system.

robertgzr:
This basically renders Dockerfile.template obsolete, since the same can
be achieved using build args. We simply pass the same values to the
docker build and let the user use them. This also includes
TARGETPLATFORM, which buildx injects:
https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images

A sample dockerfile could look like this:
```
ARG BALENA_MACHINE_NAME
FROM balenalib/${BALENA_MACHINE_NAME}-alpine

ARG BALENA_ARCH
RUN do some thing with ${BALENA_ARCH}
```

Closes https://github.com/balena-io-modules/balena-multibuild/issues/67

Change-type: minor


lib/resolve.ts
```diff
@@ -108,6 +108,16 @@ export function resolveTask(
 		templateVars,
 	);
 
+	// inject arguments
+	task.args = {
+		...templateVars,
+		...{
+			BALENA_ARCH: architecture,
+			BALENA_MACHINE_NAME: deviceType,
+			TARGETPLATFORM: platform,
```
dfunckt, line 117: What is `TARGETPLATFORM` and why is it not prefixed with  `BALENA_`?

That is, is it some docker-wide thing or specific to balena?

dfunckt: Never mind, I just noticed this in your description for the PR:

> This also includes TARGETPLATFORM, which buildx injects:

robertgzr: I actually just thought of `BUILDPLATFORM` which we should probably inject as well, or just drop both...

In any case this is just a nice-to-have :)

PR #101, opened 2022-04-10
[WIP] Subsume a few dependencies

dfunckt:
Merge a few dependencies into the tree to provide complete coverage for the module’s primary use case: given a docker-compose.yml, build the services, and (optionally) create a release for it at balenaCloud.

Clients that are looking to build docker-compose files typically have to depend on and import *all* of these modules, and some of the modules themselves depend on each other creating a situation where maintaining the correct versions of things is very cumbersome.

This commit is the first step to a standalone module that is reasonably useful but makes no effort to rethink the exposed APIs. Subsumed modules are exported as-is. In the future, we can improve by exporting more meaningful APIs.

The merged dependencies are:

- balena-compose-parse
- balena-release
- dockerfile-template
- resin-bundle-resolve
- resin-docker-build

Change-type: major


dfunckt:
Superseded by https://github.com/balena-io-modules/balena-compose/pull/1

